PR #10, opened 2015-08-06
Gmchanges2

gmcode:
Improve Windows support. Still WIP.
The Date should now build on Windows using VS2015 and g++/mingw with C++14 enabled.
C++11 should be sufficient for mac and Linux like platforms.


HowardHinnant:
Merged.

PR #15, opened 2015-09-27
Throw a runtime error if the install folder does not exist.

gmcode:
Replace my accidental use of NULL with nullptr.
Improve some comments.

### DIFF
--- a/tz.cpp
+++ b/tz.cpp
@@ -62,6 +62,7 @@
 
 #ifdef _WIN32
 #include <Windows.h>
+#include <io.h>
 #else
 #include <unistd.h>
 #endif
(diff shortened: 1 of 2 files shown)


gmcode:
Currently when the tzdb install path is missing, the Unix/Linux/Apple builds of this library create an empty tzdb object because missing paths/files don't trigger any error. But the non-existant path causes Windows to encounter a file system error opening it's mapping file and it throws a runtime error exception. This situation means the Linux and Windows versions do different things when the install path is missing - Linux creates an empty object and windows throws from the constructor so does not. This patch attempts to unify the behaviours by checking the path exists before attempting to load anything and if it doesn't exist throwing an runtime_error stating that fact.

Consistency in behaviour between windows and other operating systems is the the main goal of this patch, but if it's deemed desirable to create an empty object on all platforms instead of throwing, this patch could be applied but then change the throw to a return and commit that.

HowardHinnant:
Thanks!

PR #20, opened 2015-10-22
Cmake

davidhunter22:
This add CmakeLists.txt files to the project. These work for Clang and GCC at least on Ubuntu.
The visual  Studio project is generated fine and can be opened and builds the library. The tests mostly  fails though because of constexpr and the code needs some declspec stuff to export symbols to the resultant DLL


HowardHinnant:
I am against any system that requires maintaining a list of files to build.  It places too high of a burden on adding (or deleting) files (especially tests).

For now, the shell script `testit` will automatically pick up tests to run with no need for a list of tests.  It isn't a perfect test system, but it is better than an explicit list.

jkhoogland:
I agree that these explicit lists are silly, and it is easy to fix this/

I added a little bit of logic for globbing files in https://github.com/jkhoogland/date/commit/47adcf5c3476c3c7dd9079ac04db1e6c1bc7e3b0

New files are automatically picked when added to the source tree.

HowardHinnant:
Closing because it is over a year old and I haven't done anything with it (my fault).  I'm not against a CMake build system in general.  And a lot of things have changed since this P/R was made.

I don't think it is currently a good idea to build all of the tests.  The `testit` shell script is currently the best way to do that, and it purposefully doesn't pass everywhere.  But CMake might be nice to build with the tz.h lib which currently requires on macOS (as an example):

```
clang++ -std=c++14 test.cpp -I../date ../date/tz.cpp -O3 -lcurl
```

PR #21, opened 2015-10-22
Address issue 19: define NOMINMAX and WIN32_LEAN_AND_MEAN. In additio…

gmcode:
…n, fix some warnings so MSVC will compile at /W4 with no warnings.

No functional change intended.


davidhunter22:
Awesome I was about to put in an issue about the /W4 warnings!

PR #30, opened 2015-11-21
VS2013 build fixed

ivanp2015:
Hi Howard, I have finally managed to fix build of the library with Visual Stdio 2013 Update Pack 5. 
Please review and merge my changes into the mainstream version.


HowardHinnant:
Thank you for working on this!

Comments on date.h are here since the changes are too big for github to display.

*  Use of `_NOEXCEPT` is not acceptable.  This is a name reserved for the implementation.  `NOEXCEPT` would be acceptable.
*  Why has `month` been made default constructible?
*  Why has `year` been made default constructible?
*  Why has `weekday` been made default constructible?
*  Why has `weekday_indexed` been made default constructible on VS-2013 only?
*  Why has `weekday_last` been made default constructible on VS-2013 only?
*  Why has `year_month` been made default constructible on VS-2013 only?
*  Why has `month_day` been made default constructible on VS-2013 only?
*  Why has `month_day_last` been made default constructible on VS-2013 only?
*  Why has `month_weekday` been made default constructible on VS-2013 only?
*  Why has `month_weekday_last` been made default constructible on VS-2013 only?
*  Why has `year_month_day` been made default constructible on VS-2013 only?
*  Why has `year_month_weekday` been made default constructible on VS-2013 only?
*  Why has `year_month_weekday_last` been made default constructible on VS-2013 only?

* In `month_day::ok()`, static variables are not permitted in a constexpr function.
* In `year_month_day_last::day()`, static variables are not permitted in a constexpr function.
* In `year_month_day::ok()`, static variables are not permitted in a constexpr function.

* The `classify_duration` struct has changed meaning in the ways it handles durations between seconds and minutes (for example).  This code can probably be rewritten in terms of `ratio_less` such that it would be portable.

* Tab inserted on line 3614.
* I try to limit columns to 90.

* I'll have to give the `time_of_day` issue more thought.

HowardHinnant:
You've missed my comments on date.h on this page (github didn't provide a diff page to put these comments on).

ivanp2015:
Ok, I will try regexp search is VS or some other edxitpr - it supports regexp seerch.

ivanp2015:
Q:
Why has month been made default constructible?
Why has year been made default constructible?
Why has weekday been made default constructible?

A:
Becasue they are needed to implement default constructors in the rest of classes. 

Q:
Why has .... been made default constructible?
A:
The goal was preserve existing code to the maximum extent, adding changes only where VS2013 can't compile current code. So if you think default constructors can be put for other versions - ok.

ivanp2015:
Ok, see my recent changes. I've fixed formatting, default constrcutors, and `classify_duration`

HowardHinnant:
date.h review comments here as github does not display the date.h diffs (too many).

I think we should concentrate just on getting date.h ported first, and then move on to tz.h.  You've made enough progress that I am confident that we can make this happen, and your work is much appreciated!

*  We have enough diffs to worry about, so please try not to make stylistic changes (white space changes).

*  What compile-time errors are you seeing with the default constructors removed?  Not having default constructors for these types was a purposeful design decision, and I do not want to reverse it without fully understanding the issue.

* On lines 786-819, just make this one big #ifdef.  End the #ifdef (and other large #ifdef's, with
````
#endif  // !defined(_MSC_VER) || (_MSC_VER >= 1900)
````
(or whatever comment helps match up the beginning #if).

* Don't indent `save_stream` in `namespace detail`, but I agree with putting it in `namespace detail`.

*  Warning, incoming changes to namespace detail and the rounding functions.

* See comments on date.cpp on the date.cpp github diff page.

* Combine into one #ifdef lines 1493-1514.

* Example of how we should handle constants that are currently done with user-defined literals:
````
CONSTCD14
inline
bool
year_month_day::ok() const noexcept
{
    if (!(y_.ok() && m_.ok()))
        return false;
    return date::day{1} <= d_ && d_ <= (y_/m_/last).day();
}
````
If VS-2013 doesn't like `day{1}`, I know it will swallow `day(1)`, and that's ok too.  I prefer `day{1}` to `day(1)`, but not enough to introduce another #ifdef.  Code generation on a `constexpr`-aware compiler will be identical, and more efficient than the C++03 idiom with a `static`.  I don't mind sacrificing some small amounts of efficiency on the VS-2013 platform.

* Nice job on `classify_duration`.  One nitpick:  use `days::period` in place of `ratio<86400>`.

*  On `time_of_day`:  Instead of all the specializations, will VS-2013 swallow this?
````
template <class Duration>
class time_of_day
    : public detail::time_of_day_storage<Duration>
{
    using base = detail::time_of_day_storage<Duration>;
public:
#if !defined(_MSC_VER) || (_MSC_VER >= 1900)
    using base::base;
#else

    template <class ...Args,
              class = typename std::enable_if
                      <
                          std::is_constructible<base, Args...>::value
                      >>
    explicit time_of_day(Args&& ...args)
        : base(std::forward<Args>(args)...)
        {}

#endif  // !defined(_MSC_VER) || (_MSC_VER >= 1900)
};
````

ivanp2015:
The last trick doesn't work 

1>e:\prowork\public\github\date\date.h(3727): error C2660: 'date::time_of_day<std::chrono::hours>::time_of_day' : function does not take 2 arguments
1>e:\prowork\public\github\date\date.h(3735): error C2660: 'date::time_of_day<std::chrono::minutes>::time_of_day' : function does not take 3 arguments
1>e:\prowork\public\github\date\date.h(3744): error C2660: 'date::time_of_day<std::chrono::seconds>::time_of_day' : function does not take 4 arguments

So will have specializations.

ivanp2015:
Here's what I get when I'm trying to make month non-default constructible and comment out default constructor of the month_day

1>  tz.cpp
1>e:\prowork\public\github\date\date.h(471): error C2512: 'date::month' : no appropriate default constructor available
1>e:\prowork\public\github\date\date.h(496): error C2512: 'date::month' : no appropriate default constructor available
1>e:\prowork\public\github\date\date.h(521): error C2512: 'date::month' : no appropriate default constructor available
1>e:\prowork\public\github\date\date.h(549): error C2512: 'date::month' : no appropriate default constructor available
1>e:\prowork\public\github\date\date.h(668): error C2512: 'date::month' : no appropriate default constructor available
1>e:\prowork\public\github\date\date.h(733): error C2512: 'date::month' : no appropriate default constructor available
1>e:\prowork\public\github\date\tz_private.h(54): error C2512: 'date::MonthDayTime::pair' : no appropriate default constructor available
1>  date.cpp
1>e:\prowork\public\github\date\date.h(471): error C2512: 'date::month' : no appropriate default constructor available
1>e:\prowork\public\github\date\date.h(496): error C2512: 'date::month' : no appropriate default constructor available
1>e:\prowork\public\github\date\date.h(521): error C2512: 'date::month' : no appropriate default constructor available
1>e:\prowork\public\github\date\date.h(549): error C2512: 'date::month' : no appropriate default constructor available
1>e:\prowork\public\github\date\date.h(668): error C2512: 'date::month' : no appropriate default constructor available
1>e:\prowork\public\github\date\date.h(733): error C2512: 'date::month' : no appropriate default constructor available

ivanp2015:
Especially look at this
1>e:\prowork\public\github\date\tz_private.h(54): error C2512: 'date::MonthDayTime::pair' : no appropriate default constructor available

This is because in VS2013 we have to change union into struct (becase VS2013 refuses to compile that union).

ivanp2015:
ok, finally I have found a way to olve that differently, restoring your original code in date.h and doing some adjustements in the tz_private.h

ivanp2015:
Also managed to drop date.cpp but still we will have static const array in the function for VS2013 version - but for it it chnages nothing as long as #define CONSTCD14 is empty.

HowardHinnant:
I'd like to do *only* date.h as first step.  And I'd like to do it with your name on the git command, but my git-fu isn't good enough to do it without your help.  Here is what I'm proposing, which is date.h only, and has some tweaks from your latest date.h:

https://github.com/HowardHinnant/date/commit/cb72fe04bea93b837b0836f72aadacbd3f6bb504

The biggest unknown in this "tweak" is whether VS-2013 will swallow `time_of_day` or not.  You previously reported that it would not.  Please check again.  And if it doesn't, what about with removing the `is_constructible` constraint?

If these changes work for you, can you push them back here (and only date.h), and I'll merge it into master.  Then we can work on tz.

HowardHinnant:
If it makes it easier to start over with a new pull request, no problem there.

gmcode:
Just a few mostly irrelevant comments. The declaration that "Library is declared to be C++11 compatible, but uses std::quoted which is C++14" isn't correct. The comment attached to gmchanges2 which introduced VS support made it clear that initial support was for VS2015 and that C++14 features were used . Also VS doesn't support C++11 vs C++14 distinctions like gnu does so such declarations are misleading anyway as there is no -std=C++11 or C++14 flag for VS. MS supports features as it sees fit e.g. VS2015 has some C++14 features even before all C++11 features are fully supported.

I don't have VS2013 installed on my machine to make any especially certain comments regarding this patch, but I'd suggest just verifying after your changes relating to std::quoted etc. and getline continue to work properly between Linux and windows etc. as getline doesn't magically handle /skip any line endings in in a cross platform way. Also I'm not sure but TZDB needs a custom operator = (TZDB&&) function? It seems a default one should be fine if everything is simply a move?

It's a shame people are stuck using VS2013, because VS2015 is much better and Update 1 is only days away adding even more features and the Community Edition is free which date works with.

ivanp2015:
@gmcode,  all changes made, including assignment operator, are made because VS2013 doesn't compile that. I tell you the big secret - there are some project I know that still use VS2012, VS2010, even VS2005!!!!! That's the true situation that I can observe. But, the main point that these projects WORK FINE and they are in the "maintenance mode", so noone going to spend time or upgrading them actively to newer toolchain, but still rarely, in order to make required change to software, you need to search for something that works with that toolchain or backport (which i'm doing). 

So to me, VS2012 is not so outdated (compared to other variant listed above), and will work for couple of year more. Moreover, Microsoft has shipped Updated 5 this year, and the same as VS2015, VS2013 Community Edition is free. 

VS2013 has no special flags for switching active C++ standard as long as it compiles by default on the its maximum  C++ standard capability, 

Sorry, I did not have so much time to dig into project histrory, so I was rwally not aware of that VS support comes with assumptions of C++ 14. But now that still no matter, I need to have it buildable with VS2013

ivanp2015:
Hi Howard, personally to me you lib made sense because it support TZ part, so personally mostly like to donate primarily port of it to your proejct. It looks to me strange that we will be in situation that "some part works, some part not", esepecially taking into account that in overall this is not so big project. So personally  I don't my to have it taking more time, but commit everything at once, I can use my own build for my private purposes, it is good engough for me even if it has some non-conforming  formatting or extra static data, but I am ready to polish the code to have it finally released to public as my thanks to you for that library and as donation to open-source comminity.

ivanp2015:
class time_of_day - yes, really works without `is_constructible`, so I do new commit.

gmcode:
@ivanp2015 It seems curious to me what kind of project that is in "maintenance mode" but then decides to use C++11 and then add a brand new modern project like Date oh and then spend time downgrading it instead of upgrading their project! But I shouldn't tease. I know there are projects out there that move in their own circles. Good luck with the patch.

ivanp2015:
@gmcode, personally I need this project not due to its date.h capabilities, but due to tz capabilities, for the project that stick to VS2013. But tz part doesn't work w/o date.h part, so I need to port both.

HowardHinnant:
Please create a new P/R with only date.h

PR #32, opened 2015-11-29
TZ stuff fixed for VS2013

ivanp2015:
TZ part


HowardHinnant:
This looks very promising, just two issues, one of which is just an easy type-o.  I'll await fixes.  Thanks!

HowardHinnant:
Just one minor type-o to go!

PR #43, opened 2016-03-15
Fixes for move constructors and assignment operators added for VS2013

tomy2105:
I believe move constructors and assignment operators added by ivanp2015 to be able to compile under VS2013 are flawed and need to be fixed (apart from name in Zone and zones in TZ_DB other members are not initialized correctly!).


HowardHinnant:
Wow, thanks!

PR #46, opened 2016-03-31
Remove floor, ceil, round and abs and trunc when VS Update 2 is prese…

gmcode:
…nt as it's chrono has this polishing.

### DIFF
--- a/date.h
+++ b/date.h
@@ -877,6 +877,10 @@ trunc(T t) NOEXCEPT
 
 }  // detail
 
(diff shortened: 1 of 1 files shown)


HowardHinnant:
Thanks!  I put `trunc` back in because `iso_week` uses it.  Maybe we take it out later...

gmcode:
Cool thanks. I knew I'd get it wrong taking out trunc in the short term, twice I nearly left it in. lol

PR #56, opened 2016-06-03
Suggestions.

gmcode:
Possibly improve cross platform support.
Make validate work for C++11 because the library supports it.
Fix an unused variable mistake i made earlier.
Make constructor public, seems it should be. Possible compiler bug?

### DIFF
--- a/test/tz_test/validate.cpp
+++ b/test/tz_test/validate.cpp
@@ -101,14 +101,14 @@ tzmain()
     {
         std::cout << name << '\n';
         auto z = locate_zone(name);
-        auto begin = sys_days(jan/1/year::min()) + seconds(0);
-        auto end   = sys_days(jan/1/2035) + seconds(0);
+        auto begin = sys_days(jan/1/year::min()) + seconds{0};
+        auto end   = sys_days(jan/1/2035) + seconds{0};
         auto info = z->get_info(begin);
         std::cout << "Initially:           ";
-        if (info.offset >= seconds(0))
+        if (info.offset >= seconds{0})
             std::cout << '+';
         std::cout << make_time(info.offset);
-        if (info.save == minutes(0))
+        if (info.save == minutes{0})
             std::cout << " standard ";
         else
             std::cout << " daylight ";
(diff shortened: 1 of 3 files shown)


test/tz_test/validate.cpp
```diff
@@ -128,10 +128,10 @@ tzmain()
             auto ymd = year_month_day(dp);
             auto time = make_time(begin - dp);
             std::cout << ymd << ' ' << time << "Z ";
-            if (info.offset >= 0s)
+            if (info.offset >= seconds(0))
                 std::cout << '+';
             std::cout << make_time(info.offset);
-            if (info.save == 0min)
+            if (info.save == minutes(0))
```
HowardHinnant: Stylistic nitpick:  How about we use `minutes{0}` (curlies) instead?

tz.cpp
```diff
@@ -126,7 +126,7 @@ CONSTDATA auto max_day = date::dec/31;
 
 namespace detail
 {
-class undocumented {explicit undocumented() = default;};
+class undocumented {public: explicit undocumented() = default;};
```
HowardHinnant: Wow, I don't know how that compiled.  Instead of putting public in there, lets change it to a `struct`.

tz.h
```diff
@@ -1431,11 +1432,18 @@ parse(std::basic_istream<CharT, Traits>& is,
...
+#else
+                auto tt = timegm(&tm);
+#endif
+#if 0
 #ifndef _MSC_VER
                 auto tt = timegm(&tm);
 #else
                 auto tt = _mkgmtime(&tm);
 #endif
+#endif
```
HowardHinnant: Ok, but remove the `if 0` part.

tz.cpp
rhalbersma: https://groups.google.com/a/isocpp.org/forum/?fromgroups#!topic/std-discussion/ShmXFZg-ikQ

serendipity?

HowardHinnant:
Thanks, merged.

PR #61, opened 2016-06-20
Check if mapping file exists when !AUTO_DOWNLOAD

nico-engels:
Hello, I'm using the library in some projects and it's working very well =]... On tz support in win, I use zoned functions locally (-DHAS_REMOTE_API=0) and the [commit of 16_d/jun](https://github.com/HowardHinnant/date/commit/55087c6cbdf156a69c57fd9935c567eefc4232a7) changes the mapping using TinyXml2 and the xml from unicode.org, but the csv is still there... shouldn't be removed? 

This PR do this and check the existence of windowsZones.xml file that segfaulted on my machine when I updated the library.


gmcode:
Hi Nico

Thanks for your pull request, that's a good idea. I recommend resubmitting it with this small change:

Change the load_zone_mappings_from_xml_file function in tz.cpp  and do this:

    tinyxml2::XMLError result = doc.LoadFile(input_path.c_str());
    if (result != tinyxml2::XML_SUCCESS)
    {
        std::string msg = "Timezone mapping file not found at \"";  
        msg += input_path;  
        msg += '\"';
        throw std::runtime_error(msg);  
    }

I think it will be more robust and still do what you rightly recommend.
Thanks for the catch.

nico-engels:
Sorry, I did not followed the flow, I thought the XML library do not check the existence of the file. Is this OK, now?

Thanks for your work and patience!

gmcode:
Hi Nico,

tinyxml does this in it's LoadFile function:
    FILE* fp = callfopen( filename, "rb" );
    if ( !fp ) {
        SetError( XML_ERROR_FILE_NOT_FOUND, filename, 0 );
        return _errorID;
    }

So it returns XML_ERROR_FILE_NOT_FOUND to us if the file does not exist and in my code to you in my first message we will respond by throwing an exception unless I am mistaken, but I tested it and it seemed to work for me, but let me know if it doesn't work for you.

nico-engels:
Hello, yes it worked for me... I had also changed the message to: 
'Could not load the timezone mapping file at "blah ..." '

To cope with others errors, So I think is ok to test to XML_SUCCESS == XML_NO_ERROR (0).

HowardHinnant:
Merged.

PR #62, opened 2016-06-21
Allow the option to not use the std::system api and make the default …

gmcode:
…NOT use it for all platforms.


HowardHinnant:
Merged.

PR #64, opened 2016-06-24
Remove tinyxml2 dependency. Implement basic xml parser just sufficien…

gmcode:
…t to parse mappings file.

### DIFF
--- a/tz.cpp
+++ b/tz.cpp
@@ -84,7 +84,6 @@
 #include <memory>
 #include <sstream>
 #include <string>
-#include <tuple>
 #include <vector>
 #include <sys/stat.h>
 
@@ -123,13 +122,7 @@
 #include <curl/curl.h>
 #endif
 
-#if TIMEZONE_MAPPING
-// See comments in tz.h regarding the XML mapping file.
-#include "tinyxml2.h"
-#endif
-
 #if _WIN32
-
 static CONSTDATA char folder_delimiter = '\\';
 
 namespace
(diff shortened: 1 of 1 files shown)


tz.cpp
```diff
@@ -352,36 +345,164 @@ get_download_mapping_file(const std::string& version)
...
 static
-std::vector<date::detail::timezone_mapping>
-load_zone_mappings_from_xml_file(const std::string& input_path)
+std::vector<detail::timezone_mapping>
+load_timezone_mappings_from_xml_file(const std::string& input_path)
 {
-    using tinyxml2::XMLDocument;
-    std::vector<date::detail::timezone_mapping> zone_map_list;
-    XMLDocument doc;
+    size_t line_num = 0;
```
HowardHinnant: `std::size_t`

tz.cpp
```diff
@@ -352,36 +345,164 @@ get_download_mapping_file(const std::string& version)
...
         msg += input_path;
-        msg += '\"';
+        msg += "\".";
         throw std::runtime_error(msg);
     }
 
-    auto supplementalData = doc.FirstChildElement("supplementalData");
-    auto windowsZones = supplementalData->FirstChildElement("windowsZones");
-    auto mapTimeZones = windowsZones->FirstChildElement("mapTimezones");
+    auto error = [&](const char* info)
```
HowardHinnant: I would prefer explicit captures:
````
 auto error = [&input_path, &line_num](const char* info)
````
Rationale:  Effective Modern C++ Item 31is pretty convincing.

File: tz.cpp [hunk 3]
```diff
@@ -352,36 +345,164 @@ get_download_mapping_file(const std::string& version)
...
+        std::string msg = "Error loading time zone mapping file \"";
+        msg += input_path;
+        msg += "\" at line ";
+        msg += std::to_string(line_num);
+        msg += ": ";
+        msg += info;
+        throw std::runtime_error(msg);
+    };
+    // [optional space]a="b"
+    auto read_attribute = [&](const char* name, std::string& value, size_t startPos)->size_t
```
HowardHinnant: Ditto on the explicit captures here.

tz.cpp
```diff
@@ -352,36 +345,164 @@ get_download_mapping_file(const std::string& version)
...
+    // Quick but not overly forgiving XML mapping file processing.
+    bool mapTimezonesOpenTagFound = false;
+    bool mapTimezonesCloseTagFound = false;
+    bool mapZoneOpenTagFound = false;
+    bool mapTZoneCloseTagFound = false;
+    size_t mapZonePos = std::string::npos;
+    size_t mapTimezonesPos = std::string::npos;
+    static char mapTimeZonesOpeningTag[] = "<mapTimezones ";
+    static char mapZoneOpeningTag[] = "<mapZone ";
+    size_t mapZoneOpeningTagLen = sizeof(mapZoneOpeningTag) / sizeof(mapZoneOpeningTag[0]) - 1;
```
HowardHinnant: I think `CONSTDATA` would be a good qualifier for the above three variables, and then you could drop the `static` on the first two.
````
CONSTDATA char mapTimeZonesOpeningTag[] = "<mapTimezones ";
// etc.
````

HowardHinnant:
Overall looks awesome, thanks!

gmcode:
Hi Howard, all your changes are addressed in a new pull #65.
I started the git process again, wrestling with git is a job in of itself. Sorry if this causes anyone any trouble.

PR #66, opened 2016-07-03
Prevent warnings when using -Wundef flag

ravenvz:
I had issues with clang 3.9 that does not define _MSC_FULL_VER on linux (at least by default).
As _MSC_FULL_VER is Microsoft-specific predefined macros, I think that it is reasonable to check
if it is defined first.
This fix forces all compilers that are not defining this macro to take the first path and that might not be the perfect solution.

### DIFF
--- a/date.h
+++ b/date.h
@@ -957,7 +957,7 @@ trunc(const std::chrono::duration<Rep, Period>& d)
 }
 
 // VS Update 2 provides floor, ceil, round, abs in chrono.
-#if _MSC_FULL_VER < 190023918
+#if !defined(_MSC_FULL_VER) || _MSC_FULL_VER < 190023918
 
 // round down
 template <class To, class Rep, class Period>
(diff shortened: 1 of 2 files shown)


HowardHinnant:
What was the issue?  If `_MSC_FULL_VER` is not defined `#if _MSC_FULL_VER < 190023918` should be equivalent to `#if 0 < 190023918`.

ravenvz:
I'm sorry, I should had mentioned this earlier - I'm compiling with <b>-Wundef</b> and <b>-Werror</b>, cause I don't like undefined macros in my code. If no one else is bothered by that, then just ignore that.

HowardHinnant:
Thanks.  I would like to see this fixed.  When compiling with `-Wundef` using clang I'm getting another hit at:
````
date.h:3832:7: warning: '_MSC_VER'
````
And then tz.h has another boatload.

Would you like to tackle all these or would you prefer that I fix them?  Either way is good by me.  Just if we're going to address compiling with `-Wundef`, I want to make sure the job is complete.

Thanks for your feedback.

ravenvz:
Indeed, I missed one at line 3832. 
I haven't yet looked at tz.h, as I don't use it, but sure, I'll fix them both, probably later today.

HowardHinnant:
Thanks, merged.

PR #67, opened 2016-07-19
Remove support for using system() now other means seem acceptable. So…

gmcode:
…me other small suggestions.

### DIFF
--- a/test/tz_test/validate.cpp
+++ b/test/tz_test/validate.cpp
@@ -90,9 +90,9 @@ tzmain()
     auto& db = get_tzdb();
     std::vector<std::string> names;
     names.reserve(db.zones.size() + db.links.size());
-    for (auto& zone : db.zones)
+    for (const auto& zone : db.zones)
         names.push_back(zone.name());
-    for (auto& link : db.links)
+    for (const auto& link : db.links)
         names.push_back(link.name());
     std::sort(names.begin(), names.end());
     std::cout << db.version << "\n\n";
--- a/tz.cpp
+++ b/tz.cpp
@@ -432,8 +432,6 @@ load_timezone_mappings_from_xml_file(const std::string& input_path)
     // Quick but not overly forgiving XML mapping file processing.
     bool mapTimezonesOpenTagFound = false;
     bool mapTimezonesCloseTagFound = false;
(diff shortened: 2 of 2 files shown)


test/tz_test/validate.cpp
```diff
@@ -90,9 +90,9 @@ tzmain()
     auto& db = get_tzdb();
     std::vector<std::string> names;
     names.reserve(db.zones.size() + db.links.size());
-    for (auto& zone : db.zones)
+    for (const auto& zone : db.zones)
```
nabijaczleweli: `auto&&` would be better

gmcode: Thanks for your comments. Why do you suggest that && is better? There is no intent to move or modify the elements being iterated so I would (and currently still do) think const auto& is better than &&. as it conveys that. This article:
http://stackoverflow.com/questions/15927033/what-is-the-correct-way-of-using-c11s-range-based-for
if I read it correctly seems to conclude the same opinion as me - note at the end comments about if "observing" prefer const auto&

If you think that article is wrong or I have read it incorrectly do point me to the correct reference. Thanks!

nabijaczleweli: `auto&&` a universal reference and will be `const T&` if the container is `const` and non-const otherwise. It's more concise and makes the type less explicit.

gmcode: But I want explicitness because I always want my reference to be const to explicitly show I am not modifying any objects in the loop even if the container would let me. Doing that adds to readability and clarity imho. Conciseness is nothing if it introduces ambiguity. Your comments contradict that idea. Why does that article not convince you?

I also note that the C++ committee rejected an even more "concise" for (e : c) syntax proposal, that was short hand exactly for your for (auto&& e : c) suggestion and I suspect they rejected that proposal exactly for the reason I am suggesting - i.e. people would  be encouraged to use it because it is "concise" but then fail to convey their intent not to modify the container. So I suspect the Committee too decided clarity is more important than being concise here. Have I have convinced you? If not, perhaps you can link something to support your ideas more or explain what it is in the article I linked or that I have said that you find unconvincing?

HowardHinnant:
Merged.

PR #68, opened 2016-08-03
Fix c++11 msl and get_units overloads return expression

xaxousis:
The implementation of these functions used the expression:
```c++
    return {1, 'x'};
```
This call the initializer list constructor of basic_string, which lead
to two char strings.

This changes the expression to:
```c++
    return {'x'};
```

### DIFF
--- a/chrono_io.h
+++ b/chrono_io.h
@@ -465,7 +465,7 @@ inline
 std::basic_string<CharT>
 msl(std::atto) noexcept
 {
-    return {1, 'a'};
+    return {'a'};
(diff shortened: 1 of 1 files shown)


HowardHinnant:
On what platform are you seeing this behavior?

xaxousis:
I got this using both GCC 6.1.0 and clang 3.8 using a Debian installation.

HowardHinnant:
Thanks!  I was careless in my testing of the C++11 branch.

PR #69, opened 2016-08-13
iOS support

schmidt9:
Added support for iOS
For unpacking used https://github.com/nvh/NVHTarGzip

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,8 @@ This is actually several separate C++11/C++14 libraries:
 
 2. `"tz.h"` / `"tz.cpp"`  are a timezone library built on top of the `"date.h"` library.  This timezone library is a complete parser of the IANA timezone database.  It provides for an easy way to access all of the data in this database, using the types from `"date.h"` and `<chrono>`.  The IANA database also includes data on leap seconds, and this library provides utilities to compute with that information as well.  See http://howardhinnant.github.io/date/tz.html for more details.
 
+    There will be a talk on this library at [Cppcon 2016](https://cppcon2016.sched.org/event/7nLc/welcome-to-the-time-zone).
+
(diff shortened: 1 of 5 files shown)


HowardHinnant:
It looks to me like `wordexp` is available on iOS:
https://developer.apple.com/library/ios/documentation/System/Conceptual/ManPages_iPhoneOS/man3/wordexp.3.html

What issues did you have when running `wordexp`?

HowardHinnant:
I'm guessing that the problem you had was in specifying the correct paths using C++.  That you need to use Apple's Core Foundation Library (or whatever its called) to get the correct path.  If that's correct, I'd like to merge this.

One remaining issue:  Your copyright is too strict for this open source project.  Can you use the MIT license, such as in tz.cpp?

schmidt9:
The link you have given refers to os x, not ios, although "ios" word in link text may be confusing. XCode gives me error "'wordexp' is unavailable: not available on iOS". I'll correct copyright.

schmidt9:
The same for wordfree

HowardHinnant:
If you're having git problems correcting the copyright, I can take care of that for you.  I just didn't want to do something like that without your permission.

schmidt9:
It seems that there is no chance to get out only with c++. iOS does not allow to create folders and files in app's home directory, calling mkdir etc works on Simulator, but fails on real device. Core Foundation lets only read file system info, not write something. It is possible to change file extension with ObjC code (ios.mm) from mm to cpp to get rid of possible compiler complains, that's all.

schmidt9:
It is related to db unpacking, if we move this part to app's ObjC code and leave in cpp file only db reading logic, then with Core Foundation it is possible.

HowardHinnant:
What you have proposed above works, right?

HowardHinnant:
Merged.

PR #72, opened 2016-08-21
Namespaces, ios macro

schmidt9:


HowardHinnant:
I believe that somewhere you're going to need to `#include "TargetConditionals.h"` (`#import` if in Objc).  And then only under `#ifdef __APPLE__`.

Also you should not use `#ifdef TARGET_OS_IPHONE`, but instead `#if TARGET_OS_IPHONE`.  If I put this in a header on macOS:
````
#include "TargetConditionals.h"
````
And compile with this:
````
clang++ -std=c++14 test.cpp -E -dM
````
I get this:
````
#define TARGET_OS_IPHONE 0
````
If you would like for me to make these changes for you, I can, and I don't mind.

schmidt9:
`TargetConditionals.h` is included by Foundation/Foundation.h, which I include in pch file.

schmidt9:
Xcode complains about ObjC types, it can be solved by adding `-x objective-c++` flag in Compiler flags

schmidt9:
All this I was going to write in tz.html, as suggested by you

HowardHinnant:
`-x objective-c++` sounds like a good addition to tz.html, thanks.

Do you think it is a good idea to require your clients to use a pch supplied by the Xcode environment?  Or would it be easier if they didn't have to worry about setting up the pch?

schmidt9:
Well then we must point in some other way to `TARGET_OS_IPHONE`, particulary in `tz.cpp`. Include `TargetConditionals.h` into it?

schmidt9:
How can we point to `TARGET_OS_IPHONE` then?

schmidt9:
Like that?

img_20160822_084643

schmidt9:
Like that? Now without pch.
img_20160822_085236

schmidt9:
But in this case I get linker error

img_20160822_085601

schmidt9:
But in this case I get linker error

img_20160822_090122

HowardHinnant:
I was thinking something more like this: http://stackoverflow.com/a/5920028/576911
I.e.
````
#if __APPLE__
#  include <TargetConditionals.h>
#  if TARGET_OS_IPHONE
````
I can't see what symbols are missing in the link error you reference.  Perhaps they are resolved by including Foundation/Foundation.h when targeting iOS?

schmidt9:
Solved this way. Target_os_iphone did not want to resolve, that's why get_tzdata_path was not available 
img_20160822_184429

HowardHinnant:
That looks promising!

schmidt9:
Yes. Could you kindly make aforementioned changes in tz.cpp and ios.mm? After that maybe we could merge it finally

HowardHinnant:
Yes, I will do it this evening (I'm @ UTC -0400).

schmidt9:
Thank you

PR #75, opened 2016-08-25
Problem: date::format(..) doesn't compile

gabm:
Solution: 
- add missing locale (first parameter is always locale)
- add missing include to *sstream* since basic_ostringstream is incomplete


gabm:
Also added #include <cctype> as on vc14 the std::isdigit has its single-argument-overload there

HowardHinnant:
Thanks!

Issue #76, opened 2016-08-26
Formating of sub-systemclock-resolution timestamps


gabm:
I have two platforms with two different timestamp resolutions: Linux, which has system_clock::duration in nanoseconds and Windows, which has 100 nanoseconds. For interoperability i use nanoseconds as a common basis:

```
using my_time_point = time_point<system_clock, nanoseconds>;
```

That makes sense to me, since even though Wndows cannot *measure* this resolution, it could calculate differences between two time_points in nanoseconds anyway. Unfortunately, using the format(..) function (that now resides in date.h :smile: ) fails in this case... The problem lies here:

```
auto tt = system_clock::to_time_t(sys_time<Duration>{tp.time_since_epoch()});
```

Duration (in my case) is *nanoseconds*. On Linux thats fine, since system_clock uses nanoseconds as well. But on Windows the function to_time_t only takes *100 nanoseconds* so that one would need to cast. 

I think that one should cast the input for to_time_t to at least system_clock::duration or even seconds, since the variable is only used for timestamp down to seconds. All fraction-digits are handled before anyway, and they could be in a finer resolution than system_clock::duration... 

I'd provide a Pull Request if you agree..

HowardHinnant:
Fixed: https://github.com/HowardHinnant/date/commit/343e8299c71c7d81ea0f7bea7cf536816f82c6a2

PR #77, opened 2016-08-26
Problem: time_points with a finer resolution than system_clock::duration cannot be formated

gabm:
Soiution: cast them to system_clock::duration for the function call *to_time_t*. This shouldn't do any harm since the fractions of the second are handled maually before...


gabm:
This is the proposed solution to Issue #76

Issue #79, opened 2016-09-09
tz - with invalid (semivalid) format str, date::parse may fail to set istream in fail() condition and may return incorrect timepoints


acolomitchi:
The documentation lead me to believe the `date::parse` would accept `strftime` format strings <b>**</b>, thus is stumbled over this bug trying to use `"%FT%T%Z"` as one.<br>My initial surprise was the returned value - doesn't contains what one expected.  After reading (<i>far down on the page</i>) the spec for `date::parse`, I learned the function actually uses `time_get`-style of format string; a check on the `time_get::get()` format string spec reveals there's no `%F` (shortcut for `%Y-%m-%d`) and then my surprise changed into  "Why no sign of failure and a bad result?"

Config: `g++ (Ubuntu 5.4.1-2ubuntu1~14.04) 5.4.1 20160904`

Code to repro:

<pre>
#include &lt;iostream>
#include &lt;locale>
#include "date/tz.h"

int main() {
  namespace d=date;
  std::istringstream input;
  std::string toParse, fmt;

  std::cout << std::endl << "date::parse with invalid format (failure expected)" << std::endl;
  toParse="2016-09-09T08:45:05Z+1000";
  fmt="%FT%T%Z";
  input.str(toParse);
  input.clear();
  d::sys_time<std::chrono::milliseconds> parsedDate;
  d::parse(input, fmt.c_str(), parsedDate);
  std::cout << "Using input: `" << toParse << "` with a format of `"
            << fmt << "` (failure expected)" << std::endl
  ;
  if(input.fail()) {
    std::cout << "Failed date::parse with input date:`"<< toParse <<"`"
        << "` using format `" <<  fmt << "`"
        << std::endl
    ;
  }
  else {
    std::cout << "Parse success; applying date::make_zoned gets `"
              <<  d::make_zoned("Australia/Melbourne", parsedDate)
              <<  "`" << std::endl;
  }
  // Bug already demonstrated by the code above
  // Below is an attempt to see if my expectations (of having istream.fail()-ing)
  // are unreasonable.
  // So...
  // how does time_get behaves?
  toParse="2016-09-09T08:45:05"; // doh. Of course...
  fmt="%FT%T";                   // ...  there's no timezone
  std::cout << "\n---\nAttempting the behaviour of time_get with invalid format `"
            <<  fmt << "` (expecting failure) " << std::endl
  ;
  input.str(toParse);
  input.clear();
[...]

PR #80, opened 2016-09-09
Fix for issue #79.

acolomitchi:
Essentially:

- additionally breaking the `for` cycle at the first `err & ios_base::failbit`

- breaking the `case 'T'` if any of the delegation calls to the `f.get(...)` raises the ` ios_base::failbit` in `err`. Together with the additional cycle condition above, this should fix the bug.

I'm not sure though if this is enough: the `EOF` or `bad stream` conditions doesn't seem to be treated in the whole code of the affected function - maybe the tests should be expressed in terms of `ios_base::goodbit` and the entire processing abandoned as soon as the goodbit is zero?

### DIFF
--- a/date.h
+++ b/date.h
@@ -4167,6 +4167,7 @@ parse(std::basic_istream<CharT, Traits>& is,
                 {
                     f.get(is, 0, is, err, &tm, b, i-1);
                     if( err & ios_base::failbit ) {
+                      command = modified = false;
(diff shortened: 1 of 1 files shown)


HowardHinnant:
Fixed with https://github.com/HowardHinnant/date/commit/3b3a27efe13af1f9ddb3cbecc0b562b5a84bb4d9

PR #86, opened 2016-09-16
Clang on Fedora needs <climits> for the CHAR_BIT definition.

saschazelzer:
On my Fedora 24 box, Clang 3.8 needs an additional include directive in order to compile all tests successfully.


HowardHinnant:
Thanks!

Issue #88, opened 2016-09-17
date::format should implement an ostream version for convenience/performance reasons 


acolomitchi:
Following issue #82, it would be desirable the `date::format` function gets an overload taking and returning an `ostream` parameter.

The implementation should keep into account Howard's notes on the problem

> The 2nd form also doesn't respect width and adjustfield formatters. And even if I create the 2nd form, it will still have to create ostringstream internally for %z and for %S and %T when the precision is finer than seconds (somewhat nullifying the performance argument).

acolomitchi:
I just finished testing such a solution in a fork. Will publish and trigger a PR shortly.

HowardHinnant:
This is now done with `to_stream`: https://howardhinnant.github.io/date/tz.html#to_stream

PR #89, opened 2016-09-17
Issue #88 regarding the ostream support for `date::format`

acolomitchi:
PR for issue #88 created. As the additions to address the

>  it will still have to create ostringstream internally for `%z` and for `%S` and `%T`

are non-trivial (they are supported mainly in the  `subsecs.h` file), it will take a bit to get through all the changes in the next comment)

### DIFF
--- a/date.h
+++ b/date.h
@@ -2771,7 +2771,7 @@ year_month_weekday::ok() const NOEXCEPT
         return false;
     if (wdi_.index() <= 4)
         return true;
-    auto d2 = wdi_.weekday() - date::weekday(y_/m_/1) + days((wdi_.index()-1)*7 + 1);
+    auto d2 = wdi_.weekday() - date::weekday(static_cast<sys_days>(y_/m_/1)) + days((wdi_.index()-1)*7 + 1);
     return static_cast<unsigned>(d2.count()) <= static_cast<unsigned>((y_/m_/last).day());
 }
 
@@ -3779,8 +3779,7 @@ public:
            << use_facet<numpunct<char>>(os.getloc()).decimal_point();
         os.imbue(locale{});
 
-        time_subsec_put<Period> subs_put;
-        subs_put.put(os, t.sub_s_.count());
+        time_subsec_put<Period>::put(os, t.sub_s_.count());
 
 //#if __cplusplus >= 201402
 //        CONSTDATA auto cl10 = ceil_log10(Period::den);
@@ -3958,7 +3957,7 @@ namespace detail
 
 template <class CharT, class Traits, class Duration>
 std::basic_ostream<CharT, Traits>&
-format(std::basic_ostream<CharT, Traits>& os, std::basic_string<CharT, Traits> fmt,
+direct_stream(std::basic_ostream<CharT, Traits>& os, std::basic_string<CharT, Traits> fmt,
        const local_time<Duration>& tp, const std::string* abbrev = nullptr,
        const std::chrono::seconds* offset_sec = nullptr)
 {
@@ -3991,17 +3990,16 @@ format(std::basic_ostream<CharT, Traits>& os, std::basic_string<CharT, Traits> f
             break;
         case 'S':
         case 'T':
-            if (command && !modified && ratio_less<typename Duration::period,
-                                                   ratio<1>>::value)
-            {
+            // Wh... Aaahh, I see - you assume any ratio over second will
+            // not offer decimal places. Not true, think of ratios like 5/3
+//            if (command && !modified && ratio_less<typename Duration::period,
+//                                                   ratio<1>>::value)
+            { // the block can stay as a nest for this local type alias
                 using subsecput_type = time_subsec_put<typename Duration::period>;
-                if(subsecput_type::admits_subsecs) {
+                if(command && !modified && subsecput_type::admits_subsecs) {
                   // skip over the T or S
                   ++i;
-                  Duration subsecs=tp - floor<seconds>(tp);
-                  subsecput_type subsecs_put;
-                  // will reposition `i` as well
-                  subsecs_put.insert(fmt, i, subsecs, loc);
+                  subsecput_type::insert(fmt, i, tp, loc);
                 }
             }
             command = false;
@@ -4074,7 +4072,7 @@ format(const std::locale& loc, std::basic_string<CharT, Traits> fmt,
 {
   std::basic_ostringstream<CharT, Traits> os;
   os.imbue(loc);
-  format(os, fmt, tp, abbrev, offset_sec);
+  direct_stream(os, fmt, tp, abbrev, offset_sec);
   return os.str();
 }
 
@@ -4168,23 +4166,23 @@ format(const CharT* fmt, const sys_time<Duration>& tp)
 template <class CharT, class Traits, class Duration>
 inline
 std::basic_ostream<CharT, Traits>&
-format(std::basic_ostream<CharT, Traits>& os,
+direct_stream(std::basic_ostream<CharT, Traits>& os,
        std::basic_string<CharT, Traits> fmt,
        const local_time<Duration>& tp)
 {
-    return detail::format(os, std::move(fmt), tp);
+    return detail::direct_stream(os, std::move(fmt), tp);
 }
 
 
 template <class CharT, class Traits, class Duration>
 inline
 std::basic_ostream<CharT, Traits>&
-format(std::basic_ostream<CharT, Traits>& os,
+direct_stream(std::basic_ostream<CharT, Traits>& os,
        std::basic_string<CharT, Traits> fmt, const sys_time<Duration>& tp)
 {
     const std::string abbrev("UTC");
     CONSTDATA std::chrono::seconds offset{0};
-    return detail::format(os, std::move(fmt),
+    return detail::direct_stream(os, std::move(fmt),
                           local_time<Duration>{tp.time_since_epoch()}, &abbrev, &offset);
 }
 
@@ -4194,21 +4192,21 @@ format(std::basic_ostream<CharT, Traits>& os,
 template <class CharT, class Traits, class Duration>
 inline
 std::basic_ostream<CharT, Traits>&
-format(std::basic_ostream<CharT, Traits>& os,
+direct_stream(std::basic_ostream<CharT, Traits>& os,
        const CharT* fmt, const local_time<Duration>& tp)
 {
-    return detail::format(os, std::basic_string<CharT>(fmt), tp);
+    return detail::direct_stream(os, std::basic_string<CharT>(fmt), tp);
 }
 
 template <class CharT, class Traits, class Duration>
 inline
 std::basic_ostream<CharT, Traits>&
-format(std::basic_ostream<CharT, Traits>& os,
+direct_stream(std::basic_ostream<CharT, Traits>& os,
        const CharT* fmt, const sys_time<Duration>& tp)
 {
     const std::string abbrev("UTC");
     CONSTDATA std::chrono::seconds offset{0};
-    return detail::format(os, std::basic_string<CharT>(fmt),
+    return detail::direct_stream(os, std::basic_string<CharT>(fmt),
                           local_time<Duration>{tp.time_since_epoch()},
                           &abbrev, &offset);
 }
--- a/date.h
+++ b/date.h
@@ -47,6 +47,8 @@
 #include <utility>
 #include <type_traits>
 
+#include "subsecs.h"
+
 namespace date
 {
 
@@ -2769,7 +2771,7 @@ year_month_weekday::ok() const NOEXCEPT
         return false;
     if (wdi_.index() <= 4)
         return true;
-    auto d2 = wdi_.weekday() - date::weekday(static_cast<sys_days>(y_/m_/1)) + days((wdi_.index()-1)*7 + 1);
+    auto d2 = wdi_.weekday() - date::weekday(y_/m_/1) + days((wdi_.index()-1)*7 + 1);
     return static_cast<unsigned>(d2.count()) <= static_cast<unsigned>((y_/m_/last).day());
 }
 
@@ -3776,20 +3778,24 @@ public:
         os << std::abs(t.s_.count())
            << use_facet<numpunct<char>>(os.getloc()).decimal_point();
         os.imbue(locale{});
-#if __cplusplus >= 201402
-        CONSTDATA auto cl10 = ceil_log10(Period::den);
-        using scale = std::ratio_multiply<Period, std::ratio<pow10(cl10)>>;
-        os.width(cl10);
-        os << std::abs(t.sub_s_.count()) * scale::num / scale::den;
-#else  // __cplusplus >= 201402
-        // inefficient sub-optimal run-time mess, but gets the job done
-        const unsigned long long cl10 =
-            static_cast<unsigned long long>(std::ceil(log10(Period::den)));
-        const auto p10 = std::pow(10., cl10);
-        os.width(cl10);
-        os << static_cast<unsigned long long>(std::abs(t.sub_s_.count())
-                                              * Period::num * p10 / Period::den);
-#endif  // __cplusplus >= 201402
+
+        time_subsec_put<Period> subs_put;
+        subs_put.put(os, t.sub_s_.count());
+
+//#if __cplusplus >= 201402
+//        CONSTDATA auto cl10 = ceil_log10(Period::den);
+//        using scale = std::ratio_multiply<Period, std::ratio<pow10(cl10)>>;
+//        os.width(cl10);
+//        os << std::abs(t.sub_s_.count()) * scale::num / scale::den;
+//#else  // __cplusplus >= 201402
+//        // inefficient sub-optimal run-time mess, but gets the job done
+//        const unsigned long long cl10 =
+//            static_cast<unsigned long long>(std::ceil(log10(Period::den)));
+//        const auto p10 = std::pow(10., cl10);
+//        os.width(cl10);
+//        os << static_cast<unsigned long long>(std::abs(t.sub_s_.count())
+//                                              * Period::num * p10 / Period::den);
+//#endif  // __cplusplus >= 201402
         switch (t.mode_)
         {
         case am:
@@ -3951,8 +3957,8 @@ namespace detail
 {
 
 template <class CharT, class Traits, class Duration>
-std::basic_string<CharT, Traits>
-format(const std::locale& loc, std::basic_string<CharT, Traits> fmt,
+std::basic_ostream<CharT, Traits>&
+format(std::basic_ostream<CharT, Traits>& os, std::basic_string<CharT, Traits> fmt,
        const local_time<Duration>& tp, const std::string* abbrev = nullptr,
        const std::chrono::seconds* offset_sec = nullptr)
 {
@@ -3965,6 +3971,10 @@ format(const std::locale& loc, std::basic_string<CharT, Traits> fmt,
 
     using namespace std;
     using namespace std::chrono;
+
+    // save the ostream settings
+    save_stream<CharT, Traits> _{ os };
+    auto loc=os.getloc();
     auto command = false;
     auto modified = false;
     for (std::size_t i = 0; i < fmt.size(); ++i)
@@ -3984,13 +3994,15 @@ format(const std::locale& loc, std::basic_string<CharT, Traits> fmt,
             if (command && !modified && ratio_less<typename Duration::period,
                                                    ratio<1>>::value)
             {
-                basic_ostringstream<CharT, Traits> os;
-                os.imbue(loc);
-                os << make_time(tp - floor<seconds>(tp));
-                auto s = os.str();
-                s.erase(0, 8);
-                fmt.insert(i+1, s);
-                i += s.size() - 1;
+                using subsecput_type = time_subsec_put<typename Duration::period>;
+                if(subsecput_type::admits_subsecs) {
+                  // skip over the T or S
+                  ++i;
+                  Duration subsecs=tp - floor<seconds>(tp);
+                  subsecput_type subsecs_put;
+                  // will reposition `i` as well
+                  subsecs_put.insert(fmt, i, subsecs, loc);
+                }
             }
             command = false;
             modified = false;
@@ -4039,7 +4051,6 @@ format(const std::locale& loc, std::basic_string<CharT, Traits> fmt,
         }
     }
     auto& f = use_facet<time_put<CharT>>(loc);
-    basic_ostringstream<CharT, Traits> os;
     auto ld = floor<days>(tp);
     auto ymd = year_month_day{ld};
     auto hms = make_time(floor<seconds>(tp - ld));
@@ -4053,7 +4064,18 @@ format(const std::locale& loc, std::basic_string<CharT, Traits> fmt,
     tm.tm_wday = static_cast<int>(static_cast<unsigned>(weekday{ld}));
     tm.tm_yday = static_cast<int>((ld - local_days(ymd.year()/1/1)).count());
     f.put(os, os, os.fill(), &tm, fmt.data(), fmt.data() + fmt.size());
-    return os.str();
+    return os;
+}
+template <class CharT, class Traits, class Duration>
+std::basic_string<CharT, Traits>
+format(const std::locale& loc, std::basic_string<CharT, Traits> fmt,
+       const local_time<Duration>& tp, const std::string* abbrev = nullptr,
+       const std::chrono::seconds* offset_sec = nullptr)
+{
+  std::basic_ostringstream<CharT, Traits> os;
+  os.imbue(loc);
+  format(os, fmt, tp, abbrev, offset_sec);
+  return os.str();
 }
 
 }  // namespace detail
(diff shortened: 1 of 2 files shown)


acolomitchi:
**Preliminaries:** The `format` function works by delegating most of the work to `std::time_put` after computing the members and filling in a `struct tm` for this purpose. The author of the `date` library chose to 'intercept and patch' the provided format string for treating sub-seconds values, so that any fractional part of the seconds is 'preformated-as-value' before the call into `std::time_put`<br>The 'fmt string patching' used am essentially-runtime technique (pure runtime for C++11), creating an `osrringstream` on the stack (and corresp. heap content), 'computing the fmt string patch' using this stream and inserting the resulted patch in the appropriate place (`string::insert`).<br>The current PR re-implements the 'patching' by compile 'template only compile-time loop unrolling', so that the (assumed optimized) binary code is fully prepared to deal with the formatting of the subsecond values  without the creation of any new objects on the stack/heap.

As the amount of new code is non-trivial (about 230 CLOC), the main implementation is isolated in `subsecs.h` header file

---

(continuing)

acolomitchi:
Highlights on the changes in the `date.h` file:

1. the `subsecs.h` is `#include`d in the `date.h` even if it declares all the functionality in the `date::detail` ns and makes use of the `CONSTDATA` and related macros (as such, the `subsecs.h` is copy/pastable into the `date.h` file without effort). The reason for separating it as such: to make easier to understand the contained code (and perhaps, if the PR is accepted, to apply code formatting/styling before)

2. the new (base) method is<pre>
template &lt;class CharT, class Traits, class Duration>
std::basic_ostream&lt;CharT, Traits>&
format(std::basic_ostream&lt;CharT, Traits>& os, std::basic_string<CharT, Traits> fmt,
       const local_time&lt;Duration>& tp, const std::string* abbrev = nullptr,
       const std::chrono::seconds* offset_sec = nullptr)</pre> occupying approx the same line range as the former `std::string` overload (I actually adjusted the former code)<br>Important to be said: the `ostream` form makes use of `save_stream<CharT, Traits>`: I assumed that if that RAII object does enough in other sections of the code, it should also be sufficient here; maybe natural but still an assumption

3. the `std::string` form of the `format` function was changed to route through the `ostream` form - by making use of an `ostringstream` (as such, the net performance of the `std::string`-form of the function should be unaffected - an `ostringstream` instance creation was removed from the base function and then added back :); whatever minor, the net should however be visible in the `ostream` version)

4. the `class time_of_day_storage` friend <pre>template&lt;class CharT, class Traits>
    friend
    std::basic_ostream&lt;CharT, Traits>&
    operator<<(std::basic_ostream&lt;CharT, Traits>& os, const time_of_day_storage& t)</pre> was modified to use the `POW10` and `CEIL_LOG10` macros in the 'subsecs.h` (oh, shuck!!! And I just undef-ed them before committing - oh, well, will do another commit)/ The old code is still present though, just commented out

5. the current changes was tested with the following snippet of code:
<pre>
#include "date.h"
int main()
{
  namespace d=date;
  using namespace std::chrono;

  std::string fmt(">%F>T<%T<");
  using decis=duration&lt;std::uintmax_t, std::ratio<1,10>>;
  using millis=milliseconds;
  using micros=microseconds;
  using nanos=nanoseconds;
  using oh_one_nano=duration&lt;std::uintmax_t, std::ratio<1,10000000000>>;

  d::local_time&lt;nanos> res;

  std::istringstream in(">2016-09-17>T<12:34:56.0123456789<");
  d::parse(in, fmt, res);

  std::ostringstream dest;

  d::format(dest, fmt, d::round&lt;seconds>(res));
  // no decimal point, no nothing
  assert(dest.str()==">2016-09-17>T<12:34:56<" && !dest.bad());

  // lesser precision
  dest.str("");
  d::format(dest, fmt, d::round&lt;decis>(res));
  assert(dest.str()==">2016-09-17>T<12:34:56.0<" && !dest.bad());

  // medium precision affected by rounding
  dest.str("");
  d::format(dest, fmt, d::round&lt;micros>(res));
  assert(dest.str()==">2016-09-17>T<12:34:56.012346<" && !dest.bad());

  // higher precision than available, rounded at parsing
  dest.str("");
  d::format(dest, fmt, d::round&lt;oh_one_nano>(res));
  assert(dest.str()==">2016-09-17>T<12:34:56.0123456790<" && !dest.bad());

  // install a locale using comma as the decimal separator
  dest.imbue(std::locale("fr_FR.utf8"));
  dest.str("");
  d::format(dest, fmt, d::round&lt;millis>(res));
  assert(dest.str()==">2016-09-17>T<12:34:56,012<" && !dest.bad());

  return 0;
}
</pre>

subsecs.h
```diff
@@ -0,0 +1,282 @@
...
+ * subsecs.h
+ *
+ *  Created on: 16 Sep 2016
+ *      Author: acolomitchi
+ */
+
+#ifndef SUBSECS_H_
+#define SUBSECS_H_
+
+#include <iostream>
```
HowardHinnant: Should be `ostream`.

subsecs.h
```diff
@@ -0,0 +1,282 @@
...
+  };
+
+  static CONSTDATA unsigned value=
+      v_enable_if<
+          (base>1), // don't accept base of 0 or 1, they don't make sense
+          unsigned,
+          nl_len_tag<(val>base), val, base>::value
+      >::value
+  ;
+};
```
HowardHinnant: Everything in here should be private except `value`.

HowardHinnant: I like that this will work in C++11 whereas what it replaces requires C++14.

subsecs.h
```diff
@@ -0,0 +1,282 @@
...
+
+
+template <std::uintmax_t base>
+struct pow_struct<0,base>
+{
+  static CONSTDATA long long value=1;
+};
+
+#define POW10(width) (pow_struct<(width), 10>::value)
+#define CEIL_LOG10(val) (ceil_log_struct<(val),10>::value)
```
HowardHinnant: Since these don't have to be general purpose utilities, we might want to drop these macros and hard-code the constant `10` into the structs.

subsecs.h
```diff
@@ -0,0 +1,282 @@
...
+  std::basic_ostream<CharT>&
+  put(std::basic_ostream<CharT>& os, type value)
+  {
+    value=condition_value(value);
+    os << static_cast<CharT>('0'+(value / div_val));
+    // unroll one step further
+    decimals_fmt<width-1, T> unrolled; // the compiler MUST evaluate this
+    return unrolled.put(os, value % div_val);
+  }
+};
```
HowardHinnant: I believe we can make everything but `put` private.

subsecs.h
```diff
@@ -0,0 +1,282 @@
...
+      // insert the decimal dot
+      os << std::use_facet<std::numpunct<CharT>>(os.getloc()).decimal_point();
+      // prepare the filler
+      decimals_fmt<precision, Rep> filler;
+      // echo the value
+      filler.put(os, count);
+    }
+    return os;
+  }
+};
```
HowardHinnant: Can probably make more parts of this private.

subsecs.h
```diff
@@ -0,0 +1,282 @@
...
+
+} // namespace detail
+} // namespace date
+
+
+#undef CONSTDATA
+#undef CONSTCD11
+#undef CONSTCD14
+#undef NOEXCEPT
+
```
HowardHinnant: I like that these work in C++11.  There's some solid work in here.  I don't like the separate header.  I'd like to fold this into "date.h".

Technically your copyright on this file is too strict (not open source).  I'll need your permission to put this code (or derivatives of it) into date.h which has the MIT open source license (and add your name to the copyright holders in that file).

date.h
```diff
@@ -2769,7 +2771,7 @@ year_month_weekday::ok() const NOEXCEPT
         return false;
     if (wdi_.index() <= 4)
         return true;
-    auto d2 = wdi_.weekday() - date::weekday(static_cast<sys_days>(y_/m_/1)) + days((wdi_.index()-1)*7 + 1);
+    auto d2 = wdi_.weekday() - date::weekday(y_/m_/1) + days((wdi_.index()-1)*7 + 1);
```
HowardHinnant: This undoes a compatibility fix for VS-2013.

date.h
```diff
@@ -3776,20 +3778,24 @@ class time_of_day_storage<std::chrono::duration<Rep, Period>, detail::classify::
...
-        const unsigned long long cl10 =
-            static_cast<unsigned long long>(std::ceil(log10(Period::den)));
-        const auto p10 = std::pow(10., cl10);
-        os.width(cl10);
-        os << static_cast<unsigned long long>(std::abs(t.sub_s_.count())
-                                              * Period::num * p10 / Period::den);
-#endif  // __cplusplus >= 201402
+
+        time_subsec_put<Period> subs_put;
+        subs_put.put(os, t.sub_s_.count());
```
HowardHinnant: Nice job on refactoring this functionality that is needed in two places.

date.h
```diff
@@ -3984,13 +3994,15 @@ format(const std::locale& loc, std::basic_string<CharT, Traits> fmt,
...
-                os << make_time(tp - floor<seconds>(tp));
-                auto s = os.str();
-                s.erase(0, 8);
-                fmt.insert(i+1, s);
-                i += s.size() - 1;
+                using subsecput_type = time_subsec_put<typename Duration::period>;
+                if(subsecput_type::admits_subsecs) {
+                  // skip over the T or S
+                  ++i;
+                  Duration subsecs=tp - floor<seconds>(tp);
```
HowardHinnant: This isn't compiling for the case that `Duration` is coarser than seconds (e.g. minutes).

date.h
```diff
@@ -4140,6 +4162,60 @@ format(const CharT* fmt, const sys_time<Duration>& tp)
...
+{
+    const std::string abbrev("UTC");
+    CONSTDATA std::chrono::seconds offset{0};
+    return detail::format(os, std::basic_string<CharT>(fmt),
+                          local_time<Duration>{tp.time_since_epoch()},
+                          &abbrev, &offset);
+}
+
+
+// --------- End with ostream as destination -------
```
HowardHinnant: You've done a really nice job with this.  But I still have some potentially show-stopper concerns:

1.  I dislike the `format(cout, ...)` syntax as opposed to `cout << format(...)` syntax.  This by itself is not a huge problem as your code could (I think) be dressed up as a manipulator as is done in `<iomanip>`, which would give it the desired syntax.

2.  Assuming we do the manipulator described in (1), the following code still won't work right:
````
auto t = local_days{sep/17/2016} + 12h + 5min;
std::cout << '|' << setw(30) << left << format("%F %T", t) << "|\n";
std::cout << '|' << setw(30) << right << format("%F %T", t) << "|\n";
````
This should output:
````
|2016-09-17 12:05:00           |
|           2016-09-17 12:05:00|
````
And I'm not fond of just saying:  If you need to respect width and alignment, use format syntax #1 and if not use format syntax #2.  I want one syntax that "just works."

`std::complex` has a similar issue, and the committee solved it by specifying that the `complex` streaming operator has to stream into a `ostringstream`, and then it streams the resulting `string`.  I thought about following that pattern exactly, but some clients will actually want the result as a `string`, and so for them it seems silly to convert the internal `ostringstream` to a `string`, stream it out to another `ostream`, and then the client would have to supply an `ostringstream` and recover the `string` out of that.

I _do_ like some of your implementation, and would like to adopt it regardless of this issue of the `format` API.

File: subsecs.h (hunk 3 above)
Comment (acolomitchi) [inline subsecs.h]: Done (another commit).

File: subsecs.h (hunk 6 above)
Comment (acolomitchi) [inline subsecs.h]: Done - another commit with MIT copyright

File: date.h (hunk 7 above)
Comment (acolomitchi) [inline date.h]: As I didn't touch this code, it is perhaps an artefact generated by me forking before the fix was applied? I have no idea how this happened.

Comment (HowardHinnant) [inline date.h]: I figured as much.  Not a big deal.

File: date.h (hunk 9 above)
Comment (acolomitchi) [inline date.h]: Ooops - I missed this one. 

I think this is fixable template-style by moving the substitution into subsecs - i.e making the `time_subsec_put::insert' do the diffing itself, perhaps in two different specializations SFINAE enabled based on `Period::den` value.

I'll have a look on it - if not today (my partner wants to get out see the nature during a rainy spring) then tomorrow for sure.

Comment (HowardHinnant) [inline date.h]: No hurry.  I'm about to hop on a plane and be completely pre-occupied with cppcon-2016 for the week.

File: date.h (hunk 10 above)
Comment (acolomitchi) [inline date.h]: Just made another commit with the MIT copyright inserted - please feel free to use any parts.

In regards with the syntax, my arguments:

1. the way I see, the matter does have something related to the function name and the expectations it creates
2. the `std::time_put` doesn't pretend to offer anything related to manipulators, width and alignment of the entire result - but it's still usable and used. The same goes for the venerable `printf`in which the `fprintf` variant is highly probable used underneath by binding the stream to `stdout`.

Based on the arguments above, can we settle for renaming my proposed `ostream` variant of the `date::format` to something that doesn't create expectations of 'step into the iomanip shoes'? Something that suggest a limited and strictly utilitarian purpose, the way `std::time_put` does - I don't know, something like `date::put` or even 'do_put`?

What I'm after is a function that 'streams' the content of temporal data types straight into an `ostream` at the current position with as little overhead as possible. My current needs require 'datetime', 'date' and `time' data types, which I thought I'd be able to support on top/using `date::` data structs. Even if the library in/by itself is not concerned in anything but marshalling/unmarshalling those types, for the consumers of the library it would be a huge advantage if the structure support calendaring and arithmetic operation - this is why you propose the `date` library yourself, isn't it?

I guess what I'm asking is: please have pity for the gals and guys needing un/marshalling `date` and let them have the streaming straight from (thanks for the imanip) and to streams - no matter the name of the utility function - their needs are legitimate, so perhaps the `date` lib should cater for those needs if it aspires to the status of a standard library.

> You've done a really nice job with this. 
Thanks for the kind words, they mean something for a guy which only a month ago started to look again into C++ after about 7 years of absence - that's me.

File: subsecs.h (hunk 2 above)
Comment (HowardHinnant) [inline subsecs.h]: You make a good argument about performance.  I will definitely give it more thought (I've got a long plane ride ;-)).

In the meantime, I'm detecting subtle errors in `ceil_log_struct` that I'll leave you with. :-)  I haven't figured out the correct fix yet. Though the errors are small, they undercut my confidence in the algorithm.  Here's a test case:

````
#include "subsecs.h"
#include <iostream>

int
main()
{
    using namespace date::detail;
    std::cout << ceil_log_struct<109, 10>::value << '\n';
}
````
I'm expecting `3` but I get `2`.

**Comment by acolomitchi** [comment]
Can we coordinate a bit the plan for the near future on this regards?

The plan I'm proposing for myself:

1. fix the over-seconds duration compilation error - will require a bit of refactoring the responsibility of extracting the sub-seconds and acting appropriately;
2. apply the suggested data access levels (making private things that don't need to be public)
3. maybe rename the `format` forms of the function into `put`. Perhaps trim them to only one (the most basic form)?
4. hand it over to you to wrap/expose that utility function the way you decide (and, of course, to cannibalize the `subsecs` header content - I created it separately for the sole purpose of putting the technique into evidence.

Is the above compatible with your plans?

**Comment by HowardHinnant** [comment]
Yes, that sounds great.  Please know that I'm not yet agreeing to accept it into my branch (you're free to maintain yours), and that I'm also very grateful that you're willing to do this work.  And I don't mind changing nit-picky things on my end.  You don't need to have something perfect for me to get greedy and take it. :-)  But rest assured, if I take anything (and I'm likely to), you'll be properly attributed in the source, and in the git history.

There is likely to be a significant delay on my part on the order of a week or so because of the upcoming cppcon conference.  And associated with that delay may come higher priority updates to this lib that others point out to me during the next week.  If that happens, and if the git gets to you (it does to most people), I can help with rebasing your code to the latest HEAD if that helps.  So don't worry about the git-foo, and me continuing to push changes that are unrelated to your work.

**Comment by acolomitchi** [comment]
Thanks.

 I'll try to make the changes as 'unobtrusive' (for the lack of a better term) as possible. Anything to see the streaming into text streams as efficient as possible supported directly by the `date` - it will be a pain in the ass to maintain a separate fork - the potential users of the library would be deterred if they can't use the latest `date` and I'm not keen of constantly merging them.

**Comment by acolomitchi** [comment]
Good luck with your cppcon presentation

**Comment by HowardHinnant** [comment]
Thanks.

Spit-balling:  Rename `detail::format` to `direct_stream`?  Names are hard...

**Comment by acolomitchi** [comment]
`direct_stream` sounds great to me - conveys the purpose perfectly.

> Names are hard...
Agreed. (especially if, on top of being rusty C++-wise, one is not a native English speaker).

**Comment by acolomitchi** [comment]
Hey, listen... I'm looking over the `time_put::put/do_put`. 

Would it be in 'good taste' to derive a `time_put` class that handles the subsecs? 

(if positive, that would be a medium term approach, I'm not convinced I have the bandwidth to pull it in a matter of days, but perhaps a couple of weeks may do).

**Comment by HowardHinnant** [comment]
Interesting thought, and I'm not sure.  These primitives traffic in `std::tm` which has no spot for subseconds, so that's awkward, but not insurmountable.

**Comment by acolomitchi** [comment]
On a second though, not knowing how the `time_put` relates with `locale`, deriving one (as in `is-a`) may be perilous. If relying on a `has-a` or `uses-a` - well, that's no different from what we are doing now.

I reckon wishing you now a good night is appropriate, isn't it? (I'm AEST).

**Comment by HowardHinnant** [comment]
It's embarrassing I know what and where AEST is... :-)  Thanks much!

**Comment by acolomitchi** [comment]
Done. Committed. I hope you had a good flight, Next comment will detail the changes.

**Comment by acolomitchi** [comment]
Snippet of code used for testing:

<pre>
int main()
{
  namespace d=date;
  namespace dd=date::detail;
  using namespace std::chrono;

  std::string fmt(">%F>T<%T<");
  using ten_secs=duration&lt;std::uintmax_t, std::ratio<10,1>>;
  using decis=duration&lt;std::uintmax_t, std::ratio<1,10>>;
  using millis=milliseconds;
  using micros=microseconds;
  using nanos=nanoseconds;
  using oh_one_nano=duration&lt;std::uintmax_t, std::ratio<1,10000000000>>;
  using two_pie_and_not_2pi=duration&lt;std::uintmax_t, std::ratio<710,113>>;

  d::local_time&lt;nanos> res;

  std::istringstream in(">2016-09-17>T<12:34:56.0123456789<");
  d::parse(in, fmt, res);

  std::ostringstream dest;

  auto x=res.time_since_epoch();
  // coarser precision, no decimals
  dest.str("");
  d::direct_stream(dest, fmt, d::floor&lt;ten_secs>(res));
  assert(dest.str()==">2016-09-17>T<12:34:50<" && !dest.bad());

  // Seconds precision, no decimals
  dest.str("");
  d::direct_stream(dest, fmt, d::round&lt;seconds>(res));
  // no decimal point, no nothing
  assert(dest.str()==">2016-09-17>T<12:34:56<" && !dest.bad());

  // lesser precision
  dest.str("");
  d::direct_stream(dest, fmt, d::round&lt;decis>(res));
  assert(dest.str()==">2016-09-17>T<12:34:56.0<" && !dest.bad());
//
//
  // medium precision affected by rounding
  dest.str("");
  d::direct_stream(dest, fmt, d::round&lt;micros>(res));
  assert(dest.str()==">2016-09-17>T<12:34:56.012346<" && !dest.bad());
//
  // higher precision than available, rounded at parsing
  dest.str("");
  d::direct_stream(dest, fmt, d::round&lt;oh_one_nano>(res));
  assert(dest.str()==">2016-09-17>T<12:34:56.0123456790<" && !dest.bad());

  // stability when fed with weird periods: it compiles
  // I don't dare to assert anything about correctness
  dest.str("");
  d::direct_stream(std::cout, fmt, d::round&lt;two_pie_and_not_2pi>(res));

  // install a locale using comma as the decimal separator
  dest.imbue(std::locale("fr_FR.utf8"));
  dest.str("");
  d::direct_stream(dest, fmt, d::round&lt;millis>(res));
  assert(dest.str()==">2016-09-17>T<12:34:56,012<" && !dest.bad());

  return 0;
}
</pre>

**Comment by acolomitchi** [comment]
So, changes:

1. the most... ummm... prominent(?) one in `subsecs.h` - I took a small human step over millennia and switched the license to 'ns-MIT' - still as permissive and (I believe) as exonerating as the barbaric version of MIT <sub><sup>(large grin in very small case)</sub></sup>

2. the methods in 'struct time_subsec_put` take now a `time_point` as the parameter (whereas they used to take a `duration`) and perform the `date::floor`-ing inside. This is how I got around the compilation error for over-second `period`-s.<br><sup>MERDE!</sup> I just introduced a bug in '<< (ostream&, time_of_day)' which I need to fix now!

3. the methods in `time_subsec_put` and `decimals_fmt` are now static: no state to use/save before/during/after the call, no need for instances (even if they'd be optimized away)

4. renamed  the `format(ostream& ...)` to `direct_stream(ostream& ...)`

5. reapplied that VS2015 compatibility fix, but you better check it anyway.

**Comment by HowardHinnant** [comment]
Sorry for the delay.

I remain concerned about the accuracy of `ceil_log_struct`:
````
#include "subsecs.h"
#include <iostream>

    constexpr static int ceil_log10(unsigned long long i) noexcept
    {
        --i;
        int n = 0;
        if (i >= 10000000000000000) {i /= 10000000000000000; n += 16;}
        if (i >= 100000000) {i /= 100000000; n += 8;}
        if (i >= 10000) {i /= 10000; n += 4;}
        if (i >= 100) {i /= 100; n += 2;}
        if (i >= 10) {i /= 10; n += 1;}
        if (i >= 1) {i /= 10; n += 1;}
        return n;
    }


int
main()
{
    using namespace date::detail;
    constexpr auto i = 1'099'999ull;
    std::cout << ceil_log_struct<i, 10>::value << '\n';
    std::cout << ceil_log10(i) << '\n';
}

Output:
6
7
````
I have trouble reasoning about the rest of the code built on top of `ceil_log_struct` when I am unsure what it is computing.

**Comment by acolomitchi** [comment]
Welcome back, I hope you had a good time at cppcon.

> I have trouble reasoning about the rest of the code built on top of `ceil_log_struct` when I am unsure what it is computing.

`ceil_log_struct` computes the number of numerical figures required to minimally represent the `val` argument in base `base`. E.g. in base 10, it computes the number of (decimal) digits. Incidentally, this is also the `ceil(log10(val))`, is it not?

In your example, count how many digits the  `1'099'999ul` number requires to write down and decide which of the two cout-ed values is correct.

Doing a check using `ceil(log10(1099999))` on my calculator (no code written):
```
log10(1099999) = 6.04139229 // of course, the value is just above 10^6
cel(6.04139229)=7 // well?
```

acolomitchi:
Welcome back, I hope you had a good time at cppcon.

Hang on a bit, I'll need to clarify the exact meaning for it or (failing it) implement the proper ceil_log template and deal with the changes in the code due to the change, I'll edit this comment when I'm done.

HowardHinnant:
I'm still getting the wrong answer for `999'999ull`.

acolomitchi:
Thanks.

Done - [a2d4c76](https://github.com/acolomitchi/date/commit/a2d4c76639e6fb4f1a9be14cd2204e75ba18fe39)

(better be good, seems that I'd better have some sleep before fixing bugs) Let me know if you manage to get others.

Thanks again.

acolomitchi:
Hi Howard,

just busy or having troubles with my code? (if the later, maybe I can assist)

Regards,

Adrian

HowardHinnant:
Just busy.  Last night I convinced myself that your pow and log functions are good.  I've simplified them somewhat, but haven't pushed the work back yet.  Definitely in progress, but may be another week before it gets to master.  I'll push intermediate work back on another branch so you can track it.

HowardHinnant:
Fwiw, IANA is going to break this lib any minute, and when they do I'll be scrambling to fix that asap.

acolomitchi:
>  I'll push intermediate work back on another branch so you can track it.
I can sleep well (or at least as well as ever) even if I don't track it, but I'll be happy to if you think another pair of eyes (as red as they are) may make a difference.

> Fwiw, IANA is going to break this lib any minute, and when they do I'll be scrambling to fix that asap.
S..t, not an enviable position. Good luck with that.

HowardHinnant:
work in progress: https://github.com/HowardHinnant/date/commits/acolomitchi-master

HowardHinnant:
It has been a busy couple of months.  I'm finally getting back to this.  Partially addressed in https://github.com/HowardHinnant/date/commit/1e6afbdd91d317e94e8a1805814d0e1039627b32

I am still looking at `direct_stream`.  I will likely rename it, and completely rewrite it to involve calling the `time_put` facet only for localized results.  I'm finding that the `time_put` facet is not reliable for large ranges.  The commit referenced here is a first step towards that goal.

HowardHinnant:
I believe this commit may resolve this P/R:
https://github.com/HowardHinnant/date/commit/298e9aff2db502dcbb79fe951aa42897103655fd
However this is a fairly massive commit and I don't want to close this P/R prematurely.

acolomitchi:
I'll try to find a bit of time to shake it. But if you separated the presentation concern from the precision (and accept microfortnights as tick-units), you probably addressed #91 anyway.

PR #92, opened 2016-09-18
Continuous Integration (Windows, Linux, OS X)

saschazelzer:
Not sure if you are interested in this, but this PR adds continuous integration support via Travis CI and Appveyor. It uses GCC (Linux), Clang (Linux & OS X), Visual Studio 2013, and Visual Studio 2015.

Only the Clang (Linux & OS X) builds run the full test suite. For the Linux version, I had to slightly modify the `zone` test to only check `timezone` move assignment (without nothrow) in case of using non C++11 ABI with libstdc++11 (the Travis CI infrastructure only provides Ubuntu 14.04 which does not use the newer C++11 ABI). I thought it is worth having the full test suite also running on Linux via Clang.

The modified readme contains a table with all the CI tests, you can look at it in my fork:

https://github.com/saschazelzer/date/tree/ci-support

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,53 @@
+sudo: required
+
+language: cpp
+
+addons:
+  apt:
+    sources:
+      - ubuntu-toolchain-r-test
+      - sourceline: 'deb http://apt.llvm.org/trusty/ llvm-toolchain-trusty-3.8 main'
+    packages:
+      # This is the minimum GCC compiler version needed to compile the tz test
+      - g++-4.8
+      # This is the minimum GCC compiler version needed to compile and run the
+      # full test suite with clang using libstdc++
+      - g++-6
+      - clang-3.8
+      - libcurl4-openssl-dev
+
+matrix:
+  include:
+    - dist: trusty
+      os: linux
+      compiler: gcc
+      env: COMPILER=g++-4.8 CXX_LANG=c++11 OPTIONS=-O3 TEST_ALL=0
+    - dist: trusty
+      os: linux
+      compiler: gcc
+      # The test suite does not compile with gcc 6 because of a brace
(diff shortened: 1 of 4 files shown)


HowardHinnant:
Sorry for the delay in responding.

Here is an experimental branch:

https://github.com/HowardHinnant/date/tree/ci-support-upstream

I note that I need to enable things to get this working.  I _very_ briefly looked at the docs for doing so.  One thing I was looking for that I didn't immediately see was the "undo button".  If I get into this and don't like it, how to 100% completely recover my virginity?

nabijaczleweli:
You can disable the builds (on both Travis and AppVeyor), but I'm pretty sure the logs will stay there for posterity (on Travis, at least)

saschazelzer:
Thanks for pushing the experimental branch. Yes, you would need to log in to Travis and Appveyor using your GitHub account and enable your `date` repository.

I know that you can disable the repository integration again, but I haven't seen an `undo` or `delete account` button yet. Curiously enough, this seems to require direct contact with their support team...

HowardHinnant:
I really appreciate you doing this work, but I don't think I'm going to push it to master.  My experience with Travis and Appveyor have not been problem-free, and getting in bed with them feels more like a virus to me because of the difficulty of reversing the decision.  I'm leaving the P/R open for now to allow others to comment, and to make it easier for me to reconsider my current decision.

saschazelzer:
Okay, thanks for considering it. For me it is the only viable solution to avoid breaking Visual Studio compatibility without noticing (or any other platform a developer is not actively testing). I agree that those CI services can be cumbersome to work with, but in general I found the benefits to outweigh the disadvantages (if they manifest at all).

I can of course update my fork with your master on a regular basis and check the results myself.

Issue #94, opened 2016-09-28
Timezone database did not install correctly


ksze:
I just compiled my application with `-lcurl` and default options on Linux, which means:

- `INSTALL=~/Downloads`
- `HAS_REMOTE_API=1`
- `AUTO_DOWNLOAD=1`

But for some reason it is unable to automatically install tzdata correctly.

This is the error message I get when I run my application for the first time:

```
unreachable 0
: No such file or directory
terminate called after throwing an instance of 'std::runtime_error'
  what():  Timezone database version "2016f" did not install correctly to "/home/k_sze/Downloads/tzdata"
Aborted (core dumped)
```

I can see that the application downloaded tzdata2016f.tar.gz; it also automatically created the `~/Downloads/tzdata/` directory, but the directory is empty:

```
$ ls -l ~/Downloads/tzdata2016f.tar.gz
-rw-rw-r-- 1 k_sze k_sze 313286 Sep 28 18:12 /home/k_sze /Downloads/tzdata2016f.tar.gz
$ ls -al ~/Downloads/tzdata
total 8
drwxrwxr-x 2 k_sze k_sze 4096 Sep 28 18:12 .
drwxr-xr-x 5 k_sze k_sze 4096 Sep 28 18:12 ..
```

HowardHinnant:
Sounds like problem with the untar phase.  Can you try compiling with `-DUSE_SHELL_API=1`?

HowardHinnant:
Also please update at least to the commit made a couple hours ago labeled "Get local version from NEWS instead of Makefile".  Failure to do so will lead to excessive downloading of the tz database from the IANA website.

ksze:
Ah, so the problem is that it's not finding the tar executable. Got it.

HowardHinnant:
Hoping that's the problem.  If that isn't it, try me again. I'd love for this to work out of the box without hassles such as these.

ksze:
I also tried adding something like this in tz.cpp:

```
#ifndef TAR_PATH
#define TAR_PATH "/usr/bin/tar"
#endif
```

And I can override the default tar path using a `-DTAR_PATH=\"/bin/tar\"` compiler flag. :)

HowardHinnant:
And did that get you operational?

ksze:
Yep

ksze:
I just added an option to override the path to the tar binary here: https://github.com/ksze/date/commit/37677b709cbd571d3e3f67472308b649a1025384

[...]

PR #98, opened 2016-10-23
Add alternate tar path in extract_gz_file()

robert-git:
This is a quick and dirty fix for issue #94.  As I said in the comments of issue #94, it would be better if the tar path could somehow be automatically found, but this works for now, at least on my particular system.


HowardHinnant:
I believe this concern has been met by defaulting `USE_SHELL_API` to 1.

PR #99, opened 2016-11-01
Add support for setting path to timezone database dynamically.

bullestock:
This patch adds a function set_install() if INSTALL is not defined.

This is very useful on Windows, where you cannot be sure that any "standard" path exists.

### DIFF
--- a/tz.cpp
+++ b/tz.cpp
@@ -218,7 +218,12 @@ static std::string get_install()
 
 #ifndef INSTALL
 
-static const std::string install = get_install();
+static std::string install = get_install();
+
+void set_install(std::string install_folder)
+{
+   install = install_folder;
+}
 
 #else   // INSTALL
 
--- a/tz.cpp
+++ b/tz.cpp
@@ -228,12 +228,6 @@ get_install()
 #  else
     = expand_path("~/Downloads/tzdata");
 #  endif
-}
-
-void set_install(std::string install_folder)
-{
-    install = install_folder;
-}
 
 #else   // INSTALL
 
@@ -245,6 +239,11 @@ void set_install(std::string install_folder)
 #endif  // INSTALL
 
     return install;
+}
+
+void set_install(std::string install_folder)
+{
+    install = install_folder;
 }
 
 static
--- a/tz.cpp
+++ b/tz.cpp
@@ -214,15 +214,9 @@ namespace date
 
 using namespace detail;
 
-static std::string install;
+static std::string install
 
-static
-const std::string&
-get_install()
-{
-    install
 #ifndef INSTALL
-
 #  ifdef _WIN32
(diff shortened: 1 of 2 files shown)


tz.cpp
```diff
@@ -218,7 +218,12 @@ static std::string get_install()
 
 #ifndef INSTALL
 
-static const std::string install = get_install();
+static std::string install = get_install();
+
+void set_install(std::string install_folder)
+{
+   install = install_folder;
+}
```
HowardHinnant: This is an open question that I do not yet have a good answer for (sorry).

The ability to set the `install` at run time opens up the possibility of changing the `install` after, or even during the database initialization.  This is not an insurmountable problem.  But it is a problem I have not yet addressed to my own satisfaction.

bullestock:
Yeah, I made this strictly for my own use, and I simply want to set the
path at program startup. A simple solution would be to only allow setting
the install folder once.

tz.cpp
ta1meng: I think it is up to the client caller to ensure that the install path is set before date time library functionality is invoked (including tzdata initialization). I can't imagine anyone wanting to change the path of tzdata during program execution, after tzdata has been loaded. As a library caller I'd be very happy with some thread safety documentation added here for a function meant to set the install path: https://howardhinnant.github.io/date/tz.html#Reference. Even without that documentation I would expect client callers to infer that install path has to be set before tzdata is loaded.

The restrictions we face are:

- Our executables don't know where their install paths are until runtime. Our users can install our executables wherever they wish.

- We disabled auto download functionality, which could affect performance in unpredictable ways. We can't guarantee that users have access to the built-in download folder. We also don't want recurring downloads because our product can run inside Docker images.

- Our code base has 80+ entry points that will set the install path of the running executables. We don't want to spend time figuring out which entry points are hit when, so whenever an entry point sets the product's install path, there we pass the install path into the date time libraries, so the path setting function in question here needs to be re-entrant.

HowardHinnant:
Sorry to be slow on this.  I've attempted to address this with:
https://github.com/HowardHinnant/date/commit/634b84eb609e73a6997c7e7e8ae6dbbad6446b53
I need to maintain the `install` storage as function local static, else the fixes from:
https://github.com/HowardHinnant/date/commit/64ea0a5bc37138912f81b4f0d90186f49dd5370f
are undone.

I haven't documented this yet.  If this change solves your issue and doesn't cause others problems (for a limited testing period), I'll document it.

I won't close this P/R until it is documented.

bullestock:
No need to apologize. Your version seems to work for me. Thanks for addressing this,

HowardHinnant:
Documented (finally): https://github.com/HowardHinnant/date/commit/f3b61aab48675dd13663f01aa294fc2e7e451b07

PR #109, opened 2016-12-29
Make some const adjustments. No functional change.

gmcode:


HowardHinnant:
Looked like decent changes to me.  Why closed?

gmcode:
That's what I thought but I found g++ 6.2 and 6.3 wouldn't compile it and spewed a ton of errors so I backed it out until I could look at why. VS was fine with it though. I haven't got clang installed at the moment so I couldn't see what that made of it. I will try to get a recent build of clang installed today.

I haven't tried or have earlier versions of g++ so you could try and let me know if that works.
At that point it may well be simpler to commit your own patch and that's fine by me. If not, I will try again soon.

HowardHinnant:
Ok, thanks for the careful checking.  http://melpon.org/wandbox/ may be of help here.

PR #114, opened 2017-01-07
Use CONSTDATA in function parse enabling compiler without constexpr

lohrun:
The new implementation of the function parse is containing constexpr keyword that makes older compiler unhappy. Replacing the keyword in the function by the CONSTDATA macro enables the compilation of the library on compiler without constexpr support.


HowardHinnant:
Thanks!

PR #120, opened 2017-01-19
Enable CONSTCD14 for VS2017

invexed:
- already covered in PR #110, but there appears to be a bug with the `__cplusplus` macro in VS2017 RC (see comments in linked PR for more info)

- separate MSVC macro definitions from the generic ones; MSVC always seems to be an edge case, and this structure makes it easier to accommodate new versions if/when the need arises

- remove comments about legacy VS `constexpr` support

### DIFF
--- a/date.h
+++ b/date.h
@@ -57,15 +57,22 @@ namespace date
 // Configuration |
 //---------------+
 
-// MSVC's constexpr support is still a WIP, even in VS2015.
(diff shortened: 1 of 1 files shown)


HowardHinnant:
Merged:  https://github.com/HowardHinnant/date/commit/186dbb2891fb70685fcfad209f74ed317956a590

PR #122, opened 2017-01-30
Removing static to be compiled as library

nvelozsavino:
making the function static doesn't allow to be compiled as a library when using -Werror

### DIFF
--- a/tz.cpp
+++ b/tz.cpp
@@ -190,7 +190,7 @@ get_download_folder()
 #else // !_WIN32
 
 
-static
+//static
 std::string
 expand_path(std::string path)
 {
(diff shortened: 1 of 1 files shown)


HowardHinnant:
What does the warning say?

nvelozsavino:
error: ‘std::__cxx11::string expand_path(std::__cxx11::string)’ defined but not used [-Werror=unused-function]
 expand_path(std::string path)

HowardHinnant:
Ok, thanks.  I note that `expand_path` is not used if `INSTALL` is defined.  Can you confirm that you get this error only when defining `INSTALL`, and if so, #ifdef out `expand_path` when `INSTALL` is defined?  But keep `expand_path static` so that the symbol continues to have file scope.  That would make a much appreciated P/R.

And don't forget to add your name to the copyright list at the top of the file if you are so inclined.

nvelozsavino:
You're totally right, thanks

HowardHinnant:
Merged: https://github.com/HowardHinnant/date/commit/c1034550d2a9dd923546e3439f1896fab616e6c5

PR #138, opened 2017-03-08
Added Alloc argument to template parameters to support custom allocat…

gaspardpetit:
Minimal changes required to support a custom allocator in from_stream. This should not have impact on existing code which will default to std::allocator<CharT>, but will allow specifying explicitly a custom allocator if needed.

### DIFF
--- a/date.h
+++ b/date.h
@@ -3395,10 +3395,10 @@ to_stream(std::basic_ostream<CharT, Traits>& os, const CharT* fmt,
           const fields<Duration>& fds, const std::string* abbrev = nullptr,
           const std::chrono::seconds* offset_sec = nullptr);
 
-template <class CharT, class Traits, class Duration>
+template <class CharT, class Traits, class Duration, class Alloc = std::allocator<CharT>>
 void
 from_stream(std::basic_istream<CharT, Traits>& is, const CharT* fmt,
-            fields<Duration>& fds, std::basic_string<CharT, Traits>* abbrev = nullptr,
+            fields<Duration>& fds, std::basic_string<CharT, Traits, Alloc>* abbrev = nullptr,
             std::chrono::minutes* offset = nullptr);
 
 // time_of_day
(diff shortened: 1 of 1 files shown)


HowardHinnant:
I like the direction of this PR.

I'm not convinced the `Alloc` has to be defaulted anywhere `Traits` isn't defaulted.  And I would prefer it to be applied everywhere we have a `basic_string<CharT, Traits>` for consistency.

gaspardpetit:
I agree it is not mandatory. However, since it is guessed from `abbrev` and this parameter can be null, it will force some existing call to provide the full template parameters - breaking backward compatibility.  This is not much of an issue for those using the `parse` interface since this interface, being high-level, will keep using std::allocator.  I would be ok removing the default argument if you are ok potentially breaking some existing code.

gaspardpetit:
(by the way, side question, I always thought assigning nullptr to a std::string / std::basic_string was not well defined (http://stackoverflow.com/questions/10771864/assign-a-nullptr-to-a-stdstring-is-safe)

HowardHinnant:
You make a good point about the default argument.  I will have to consider that further, but not tonight. :-)  Thanks much for your work on this.

HowardHinnant:
How do you feel about this modification of your push?  https://github.com/HowardHinnant/date/commit/37ce971c44df92a19e77a77bd848a95fc951c070
It keeps the spirit of your addition of `Alloc` and adds to it.  And does it test out ok on your platform?

gaspardpetit:
As far as I'm concerned, this is perfect. I had forgotten to update tz.h, sorry and thanks for fixing that!  Out of curiosity, why did you decide to keep default arguments there and not in date.h?

HowardHinnant:
I've attempted to keep the default args in both files, but list them only once per signature (e.g. only on the definitions, but not the forward declarations).  If you see a spot where you think I've messed that up, let me know and I'll fix it before pushing this to master.

HowardHinnant:
Merged: https://github.com/HowardHinnant/date/commit/f4292e6aca78a1f18b582debe15926ed37ce5da0

PR #140, opened 2017-03-21
Add a more generic approach to get UNIX download folder

coin-au-carre:
Download folder can differ (especially on non-english Linux distribution). 
The changes propose to stick with the download folder provided with the [xdg-user-dir](https://freedesktop.org/wiki/Software/xdg-user-dirs/) command which is available in most Linux distribution : 

- Add an exec_and_get_output function to grab the result of xdg-user-dir with popen
- Change get_download_folder() so that it uses the above function, verify whether the output is eligible. Result should always be a valid path. 
- Added copyright if this work can be useful

### DIFF
--- a/tz.cpp
+++ b/tz.cpp
@@ -2722,7 +2722,8 @@ remote_download(const std::string& version)
 #else
     // Create download folder if it does not exist on UNIX system
     auto download_folder = get_download_folder();
-    if (!file_exists(download_folder)) {
+    if (!file_exists(download_folder)) 
+    {
         make_directory(download_folder);
     }
 #endif
@@ -2751,7 +2752,6 @@ remote_install(const std::string& version)
 
     std::string install = get_install();
     auto gz_file = get_download_gz_file(version);
-
     if (file_exists(gz_file))
     {
         if (file_exists(install))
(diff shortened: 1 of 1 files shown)


HowardHinnant:
Nice idea, but this adds an unacceptable expense to initialization for those platforms that do not have  xdg-user-dir installed (which is too many).

However this issue can hopefully be handled either at compile time by the `INSTALL` macro, or at runtime by the soon-to-be-documented `void set_install(const std::string& install)` function.

I've also updated tz.cpp with your copyright.  I should've checked that on your last contribution.

coin-au-carre:
Thank you @HowardHinnant for your answer. If `xdg-user-dir` is not available then the download folder revert back to the old default one. So what is not good is the fact that we still try to launch the command on all non WIN32 platform if I understand correctly ? 
I agree that the compile time and runtime solutions are necessary.

HowardHinnant:
A client can non-intrusively use `xdg-user-dir` if desired and send the result of that to this library via `set_install`, right?

coin-au-carre:
I see your point indeed.

HowardHinnant:
I will try to get `set_install` documented within about 12 hours.  Then your use case with `xdg-user-dir` might make a nice example in the wiki on how to use `set_install`.

HowardHinnant:
`set_install` now documented: https://github.com/HowardHinnant/date/commit/f3b61aab48675dd13663f01aa294fc2e7e451b07

coin-au-carre:
Ok nice ty I will try to add an `xdg-user-dir` example in the wiki

HowardHinnant:
Awesome, love it!

PR #141, opened 2017-03-30
Ignore annoying GCC 4.9 warning -Wmissing-field-initializers

coin-au-carre:
On compilers priori to GCC 5 we get annoying warnings (see below). 

I think the following link https://gcc.gnu.org/bugzilla/show_bug.cgi?id=61489 is revelant to this bug so I just ignore this warning on specific parts of `date.h` and `tz.cpp`. 
I could also have wrapped my `date.h` and `tz.h` in my own program with `#pragma ignore` but maybe this work could be helpful.

> /opt/atlassian/pipelines/agent/build/source/common/utils/date/date.h: In instantiation of 'void date::to_stream(std::basic_ostream<_CharT, _Traits>&, const CharT*, const date::fields<Duration>&, const string*, const seconds*) [with CharT = char; Traits = std::char_traits<char>; Duration = std::chrono::duration<long int, std::ratio<1l, 1000l> >; std::string = std::basic_string<char>; std::chrono::seconds = std::chrono::duration<long int>]':
> 
> [...]
> 
> /opt/atlassian/pipelines/agent/build/source/common/utils/coin/logger.hpp:201:53:   required from 'std::string coin::_detail::Log<Level>::timestamp_thread_safe(const string&) [with coin::LogLevel Level = (coin::LogLevel)1; std::string = std::basic_string<char>]'
> /opt/atlassian/pipelines/agent/build/source/common/utils/coin/logger.hpp:248:32:   required from here
> /opt/atlassian/pipelines/agent/build/source/common/utils/date/date.h:4217:24: warning: missing initializer for member 'tm::tm_sec' [-Wmissing-field-initializers]
>                      tm = std::tm{};
>                         ^
> /opt/atlassian/pipelines/agent/build/source/common/utils/date/date.h:4217:24: warning: missing initializer for member 'tm::tm_min' [-Wmissing-field-initializers]
> /opt/atlassian/pipelines/agent/build/source/common/utils/date/date.h:4217:24: warning: missing initializer for member 'tm::tm_hour' [-Wmissing-field-initializers]
> /opt/atlassian/pipelines/agent/build/source/common/utils/date/date.h:4217:24: warning: missing initializer for member 'tm::tm_mday' [-Wmissing-field-initializers]
> 
> [...]

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 [![Join the chat at https://gitter.im/HowardHinnant/date](https://badges.gitter.im/HowardHinnant/date.svg)](https://gitter.im/HowardHinnant/date?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
 <hr/>
 
-**[Try it out on wandbox!](http://melpon.org/wandbox/permlink/HMMMIqtuG6Ia1AHY)**
+**[Try it out on wandbox!](https://wandbox.org/permlink/n6NEh9nFCWLVgm1Q)**
 
 This is actually several separate C++11/C++14 libraries:
 


HowardHinnant:
If we're going to do this, I think I would rather just wrap all of date.h and tz.cpp with this.  And tz.cpp doesn't need the pop.  That way the code is easier to read (without all the clutter).

HowardHinnant:
(hit the wrong button to submit the comment :-) )

PR #142, opened 2017-03-31
Ignore annoying GCC 4.9 warning -Wmissing-field-initializers

coin-au-carre:
Sorry to reopen this pull request from #141 I had to restart from a fresh repository (after a messup I was forced to delete the old repository). 

> If we're going to do this, I think I would rather just wrap all of date.h and tz.cpp with this. And tz.cpp doesn't need the pop. That way the code is easier to read (without all the clutter).

For the sake of lisibility, I have then wrapped all the pragma diagnostic at top and bottom of the files. The drawdown is that we could ignore some pedantic warnings now with GCC. Is that acceptable for you ?

### DIFF
--- a/date.h
+++ b/date.h
@@ -50,14 +50,12 @@
 #include <utility>
 #include <type_traits>
 
-#ifdef __GNUC__
-// GCC complains about __int128 with -pedantic or -pedantic-errors
-# pragma GCC diagnostic push
-# pragma GCC diagnostic ignored "-Wpedantic"
 # if __GNUC__ < 5
+#  pragma GCC diagnostic push
+   // GCC 4.7 Bug 50454 complains about __int128 with -pedantic or -pedantic-errors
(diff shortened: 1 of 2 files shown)


coin-au-carre:
Ok I could find a better solution. 

In tz.cpp, I have removed one your diagnostic ignore since you have "fixed" it in this [commit](https://github.com/HowardHinnant/date/commit/6a31edcb3843d4cdfe34a9c0e7a3d871a1fe38c8). 
I have moved the diagnostic ignore pedantic in the GCC < 5 case. There is no reason to put it for recent GCC versions cf [Bug 50454](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=50454). 

So now the few ignore warnings apply only for old GCC <= 4.9. That should be acceptable I guess.

HowardHinnant:
Merged: https://github.com/HowardHinnant/date/commit/c4c2550b296a8d2e45a4b5148c1ed3464c6e7938

PR #150, opened 2017-04-28
Two error handling fixes (one fixing uninitialized stack memory being accessed)

jaakristioja:
c7bc7fe fixes the a message of a `std::runtime_error` containing a trailing semicolon and space character but no following error message if the related `strerror_r` call fails.

bd09561 fixes the incompatibility with the GNU C library where `strerror_r` is instead defined in an non-portable XSI-incompatible manner. I discovered this issue because of a compiler warning about comparing the return value of `strerror_r` (a pointer) to an integer `0` on GNU platforms. Regardless of the warning, the code still compiled, and resulted in uninitialized stack memory being read from the `message` buffer:

- Before c7bc7fe:
  - if the GNU-specific `strerror_r` succeeded, the `if (strerror_r(errno, message, sizeof(message)) != 0)` branch was taken, `message` was set to the empty string (regardless of whether `strerror_r` wrote to it) and later streamed to `os`. No undefined behavior resulted, but the actual error message was ignored.
  -  if the GNU-specific `strerror_r` failed and returned `nullptr`, the conditional would not trigger and the uninitialized `message` buffer was streamed to `os`.
- After c7bc7fe:
  - if the GNU-specific `strerror_r` succeeded, the `if (strerror_r(errno, message, sizeof(message)) == 0)` condition was not triggered and `message` was not streamed to `os`. No undefined behavior resulted, but the actual error message was ignored.
  -  if the GNU-specific `strerror_r` failed and returned `nullptr`, the conditional triggered the uninitialized `message` buffer was streamed to `os`.

### DIFF
--- a/tz.cpp
+++ b/tz.cpp
@@ -2919,6 +2919,43 @@ current_zone()
 
 #else // !WIN32
 
+// +-------------------------------------------+
+// | Begin XSI-compliant strerror_r workaround |
+// +-------------------------------------------+
+//
+// At compile-time this workaround wraps both the XSI-compliant and GNU-specific
+// strerror_r functions behind a common function interface which is identical to
+// the interface of the GNU-specific strerror_r. Doing it the other way around
+// would require copying/moving the error message into the provided buffer.
+
+namespace {
+
+template <typename strerror_r_t> struct StrErrorR;
+
+template <> // Specialization for XSI-compliant strerror_r:
+struct StrErrorR<int (int, char *, ::size_t)> {
+    static char * call(int errnum, char * buffer, std::size_t buflen) noexcept
+    { return ::strerror_r(errnum, buffer, buflen) ? nullptr : buffer; }
+};
+
+template <> // Specialization (pass-through) for GNU-specific strerror_r:
+struct StrErrorR<char * (int, char *, ::size_t)> {
+    constexpr static auto const call = ::strerror_r;
+};
+
+// Note that this definition of strerror_r is in an anonymous namespace inside
+// the ::date namespace
+static inline char * strerror_r(int errnum, char * buffer, std::size_t buflen)
+        noexcept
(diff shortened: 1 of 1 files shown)


HowardHinnant:
I agree you've identified a good issue.  Thanks!

What do you think about resolving it like this? https://github.com/HowardHinnant/date/commit/88a4d1d0b9f106b655f93e822ddddc314610f63b

In a nutshell, this uses `system_error` to translate `errno` instead of `strerror_r`, and if anything fails during the building of the  message, then that error gets thrown instead.  This seems simpler to me as it pushes the portability issue on to the std::lib vendor.

jaakristioja:
I've never used `system_error` myself because it's always seemed quite obscure to me. Maybe it's due to my poor understanding, but why not just throw a `system_error(erro, system_category(), "realpath failure: errno = " + to_string(errno))` which would derive from `runtime_error` as well? 

Other than that, I like the approach and have no objections to the implementation in 88a4d1d. The less platform-specific code you have to write yourself, the better.

PS: Your exception handling code might still throw unexpected memory allocation exceptions. But that's not an easy problem to solve, and I'm probably being too pedantic again.

HowardHinnant:
I like your `system_error(erro, system_category(), "realpath failure: errno = " + to_string(errno))` even better!  Want to code that up?  And add your name to the copyright list in tz.cpp if you want (it is encouraged but not required).  You can just edit your branch and push, and it will update this P/R.  And then I'll squash it down to a single commit prior to merge.

jaakristioja:
I made it simply `system_error(errno, system_category(), "realpath() failed")` without duplicating the `errno` information passed to the constructor in the `what` string. For g++/libstdc++ 5.4.0 this makes `what()` return something like `realpath() failed: <strerror message here>`.

HowardHinnant:
Merged: https://github.com/HowardHinnant/date/commit/245452f7c0f59a9b802f293d3d13a3583dd804a6
Thanks!

PR #151, opened 2017-04-30
fix compile warning about unused get_download_gz_file()

tomba:
get_download_gz_file() is not used when HAS_REMOTE_API=0, and causes:

warning: 'std::__cxx11::string date::get_download_gz_file(const string&)' defined but not used [-Wunused-function]

Wrap get_download_gz_file() inside #ifdef to remove the warning.

### DIFF
--- a/tz.cpp
+++ b/tz.cpp
@@ -269,6 +269,7 @@ get_install()
     return ref;
 }
 
+#if HAS_REMOTE_API
 static
 std::string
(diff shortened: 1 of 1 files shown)


HowardHinnant:
Merged: https://github.com/HowardHinnant/date/commit/44f6dfc58e79759c4f7b08728e455af29eb01025
Thanks!

PR #153, opened 2017-05-07
Implement tzfiles and CMake

Erroneous1:
This implements reading /usr/share/zone_info tzfiles (issue 1). Because they don't have the same information as the IANA rules I made a couple separate classes `tzfile_zone` and `tzfile_db`. I renamed `time_zone` and `TZ_DB` to `tzrule_zone` and `tzrule_db` respectively, and made typedefs for `time_zone` and `TZ_DB` for whatever is specified by `TIMEZONE_DEFAULT` macro. I renamed `zoned_time` to `basic_zoned_time` and it takes a `class TimeZone` as a template parameter. I also moved things like `locate_zone`, `current_zone`, `get_tzdb`, etc to the _zone and _db classes and inlined static versions to use `time_zone` and `TZ_DB`. I depated separating the 2 options into different headers and implementations but it would add up to 4 files to download for timezones to work.

I modified validate.cpp for the new setup that can take `tzfile` or `tzrule` as a command line parameter and test that timezone. When I tested tzfile and tzrule I got the same output except the version line (tzfile doesn't have a version anywhere that I can find) and tzfile has a Factory timezone. 

I haven't tested on Windows so hopefully timezone mapping still works.

I also added a CMakeLists.txt for cmake users and modified the *.fail.cpp files to do static_asserts so that they could be tested with ctest. I could do this as a separate PR or just leave it off if you would prefer to keep it as is. Also on my setup parse.pass.cpp is failing an assert, but I don't think it is related to anything I did.

As for performance, running validate tzfile on my system normally takes ~2 seconds. Running validate tzrule takes ~14 seconds without having to download any files. tzfile requires no dependencies or system calls. If you need more up to date tzfiles than what is provided by OSX you can always run zic.


HowardHinnant:
Massive effort, thanks!

When I include "tz.h" I get:

````
In file included from test.cpp:2:
../date/tz.h:325:5: error: multiple overloads of 'basic_zoned_time' instantiate to the same signature 'void (const char *, const
      local_time<std::__1::chrono::duration<long long, std::__1::ratio<1, 10000000> > > &)' (aka 'void (const char *, const
      std::__1::chrono::time_point<date::local_t, std::__1::chrono::duration<long long, std::__1::ratio<1, 10000000> > > &)')
    basic_zoned_time(const char* name,        const local_time<Duration>& tp);
    ^
test.cpp:17:43: note: in instantiation of template class 'date::basic_zoned_time<std::__1::chrono::duration<long long, std::__1::ratio<1, 10000000> >,
      char>' requested here
    return format("%d/%m/%Y %I:%M:%S %p", make_zoned("Etc/GMT-2", tp));
                                          ^
../date/tz.h:323:5: note: previous declaration is here
    basic_zoned_time(const TimeZone* z,      const local_time<Duration>& tp);
    ^
````

`testit` is failing to run in directories that don't have both `.pass.cpp` and `.fail.cpp` sources in them (at least on macOS).  I haven't yet figured out a fix.

I'm a little concerned about silent ABI changes such as `version` and the valid range (which is arbitrarily limited with `TIMEZONE_FILES`).

**Comment by Erroneous1** [comment]
Ok I compiled with clang 4.0.0 and had to fix some things. I added some fixes for testit and I think that it should work.

I couldn't  get the same error you mentioned and didn't see a test.cpp in the repo. If you still have issues with this commit, could I get a copy of this file?

I included a static constexpr char version[] so it wouldn't break people's builds that reference .version, but I don't see a way to get the actual version without quering the plethora of package managers and interpreting the results or figuring it out based on the actual data.

I'm not sure about how we can do the min year. For max year we could look at the max value from EST5EDT which is Nov 1 06:00 2499.

Also it occurred to me that changing TIMEZONE_MAPPING to be disabled with -DTIMEZONE_MAPPING=0 and use `#if TIMEZONE_MAPPING` instead of `#ifdef TIMEZONE_MAPPING` might break current setups. I can change it back if you want, but this is more like HAS_REMOTE_API. Is there a way to check if a macro was defined without a value?

**Comment by HowardHinnant** [comment]
Thanks!

A macro defined without a value should have the value 1.

It is going to be some hours before I'm able to look at all this, and probably some days before I can completely digest it.  The day-job is back. ;-)  I think the first thing I want to do is to get your branch up on my GitHub as a parallel development branch so we can both contribute to it, and keep it up to date with master that way.  I'll try to get that done after I get off work this evening.

**Comment by Erroneous1** [comment]
Ah that's good to know on macros. Shouldn't be an issue then.

In the next couple of weeks my wife and I are having a baby so any contributions I have will be sparse and probably quite delayed.

**Comment by HowardHinnant** [comment]
Congratulations!!!

**Comment by Erroneous1** [comment]
Thanks!

**Comment by Erroneous1** [comment]
I didn't add any install lines to cmake, but I could. If you were to install headers, would you install them to include/date/{date.h,tz.h}? And would you install the tz library as lib/libdate{.so,.a}? Doing a google search for /usr/include/date and /usr/local/include/date doesn't show any conflicting packages. syslog-ng has a libdate.so but it is often put under lib/syslog-ng/libdate.so so it would probably be safe to do libdate.so. We could also do libdate++{.so,.a} to distinguish it.

**Comment by HowardHinnant** [comment]
So far I've just been advising people to add a `-I` to wherever they want to put the headers, and just include the single source tz.cpp into their build.  With just the single source I'm hesitant to create a complex install/build system.

**Comment by Erroneous1** [comment]
I'll work on making sure that a cmake project can download the date library to a subdirectory, `add_subdirectory` the subdirectory, then `include_directories(${DATE_INCLUDES} ${DATE_TZ_INCLUDES})` and `target_link_libraries` `${DATE_TZ_LIBS}`.

I might also work on implementing `LAZY_INIT` to keep the memory footprint down. I think I can do it using something like `std::pair<std::string,std::unique_ptr<tzfile_zone>>` to hold the zones.

File: test/testit [hunk 1]
```diff
@@ -74,23 +74,10 @@ IMPLEMENTED_PASS=0
...
-	fi
 
-	if (ls ${TEST_PREFIX}*pass.cpp > /dev/null 2>&1)
+	if (ls ${TEST_PREFIX}*fail.cpp ${TEST_PREFIX}*pass.cpp > /dev/null 2>&1)
 	then
-		for FILE in $(ls ${TEST_PREFIX}*pass.cpp); do
+		for FILE in $(ls ${TEST_PREFIX}*fail.cpp ${TEST_PREFIX}*pass.cpp); do
             if [ "$VERBOSE" ]
             then
              	echo "Running test: " $FILE
```
HowardHinnant: I needed to add a new include directive to point to `test_type_traits.h`:
````
--- a/test/testit
+++ b/test/testit
@@ -48,7 +48,7 @@ if [ -z "$CXX_LANG" ]
 then
     CXX_LANG=c++14
 fi
-OPTIONS="-std=${CXX_LANG} $OPTIONS -I$ROOT -Wall"
+OPTIONS="-std=${CXX_LANG} $OPTIONS -I$ROOT -I$ROOT/test -Wall"
 
 case $TRIPLE in
   *-*-mingw* | *-*-cygwin* | *-*-win*)
````

**Comment by HowardHinnant** [comment]
This program:
````
#include "tz.h"
#include <iostream>

int
main()
{
    using namespace std::chrono;
    using namespace date;
    std::cout << make_zoned(current_zone(), system_clock::now()) << '\n';
}
````
Compiled like this:
````
clang++ -std=c++11 test.cpp -I../date ../date/tz.cpp -lcurl
````
Results in this run-time error:
````
libc++abi.dylib: terminating with uncaught exception of type std::__1::ios_base::failure: ios_base::clear: unspecified iostream_category error
````
When compiled with:
````
clang++ -std=c++11 test.cpp -I../date ../date/tz.cpp -lcurl -DTIMEZONE_RULES=1
````
I get a link-time error:
````
Undefined symbols for architecture x86_64:
  "date::tzrule_db::current_zone()", referenced from:
      date::current_zone() in test-e50faa.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
````

**Comment by HowardHinnant** [comment]
Here's a stack trace for that unhandled exception:
````
Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib        	0x00007fffb48dcd42 __pthread_kill + 10
1   libsystem_pthread.dylib       	0x00007fffb49ca5bf pthread_kill + 90
2   libsystem_c.dylib             	0x00007fffb4842420 abort + 129
3   libc++abi.dylib               	0x00007fffb339884a abort_message + 266
4   libc++abi.dylib               	0x00007fffb33bdc37 default_terminate_handler() + 243
5   libobjc.A.dylib               	0x00007fffb3ecb713 _objc_terminate() + 124
6   libc++abi.dylib               	0x00007fffb33bad69 std::__terminate(void (*)()) + 8
7   libc++abi.dylib               	0x00007fffb33ba7de __cxa_throw + 121
8   libc++.1.dylib                	0x00007fffb3348ff4 std::__1::ios_base::clear(unsigned int) + 118
9   a.out                         	0x00000001090ea137 date::tzfile_zone::tzfile_zone(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, std::__1::basic_istream<char, std::__1::char_traits<char> >&, std::__1::vector<date::leap, std::__1::allocator<date::leap> >*) + 567 (tz.cpp:3062)
10  a.out                         	0x00000001090eec1d date::tzfile_zone::tzfile_zone(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, std::__1::basic_istream<char, std::__1::char_traits<char> >&, std::__1::vector<date::leap, std::__1::allocator<date::leap> >*) + 45 (tz.cpp:3204)
11  a.out                         	0x00000001090f4e94 date::tzfile_db::init_tzdb(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >) + 8820 (tz.cpp:3377)
12  a.out                         	0x00000001090f55fb date::tzfile_db::get_tzdb(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) + 91 (tz.cpp:3400)
13  a.out                         	0x00000001090f56a5 date::tzfile_db::locate_zone(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) + 37 (tz.cpp:3407)
14  a.out                         	0x00000001090d8aad date::tzfile_zone::locate_zone(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) + 29 (tz.h:553)
15  a.out                         	0x00000001090d8a31 date::basic_zoned_time<std::__1::chrono::duration<long long, std::__1::ratio<1l, 1000000l> >, date::tzfile_zone>::basic_zoned_time(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, std::__1::chrono::time_point<std::__1::chrono::system_clock, std::__1::chrono::duration<long long, std::__1::ratio<1l, 1000000l> > > const&) + 193 (tz.h:1334)
16  a.out                         	0x00000001090d8933 date::basic_zoned_time<std::__1::common_type<std::__1::chrono::duration<long long, std::__1::ratio<1l, 1000000l> >, std::__1::chrono::duration<long long, std::__1::ratio<1l, 1l> > >::type, date::tzfile_zone> date::make_zoned<std::__1::chrono::duration<long long, std::__1::ratio<1l, 1000000l> >, date::tzfile_zone>(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, std::__1::chrono::time_point<std::__1::chrono::system_clock, std::__1::chrono::duration<long long, std::__1::ratio<1l, 1000000l> > > const&) + 51 (tz.h:1494)
17  a.out                         	0x00000001090d87f5 main + 165 (test.cpp:9)
18  libdyld.dylib                 	0x00007fffb47ae235 start + 1
````

**Comment by HowardHinnant** [comment]
This program works if I comment out the loading of the leap second info from `tzfile_db::init_tzdb`:
````
#include "tz.h"
#include <iostream>

int
main()
{
    using namespace std::chrono;
    using namespace date;
    std::cout << make_zoned(std::string("America/New_York"), system_clock::now()) << '\n';
}
````
I can not find leap second info anywhere under `/usr/share/zoneinfo/` on macOS (which also implies iOS).

**Comment by HowardHinnant** [comment]
This program fails to compile:
````
#include "tz.h"
#include <iostream>

int
main()
{
    using namespace std::chrono;
    using namespace date;
    auto& db = get_tzdb();
    for (auto& tz : db.zones)
        std::cout << tz.name() << '\n';
}
````
Error messages:
````
test.cpp:9:16: error: use of undeclared identifier 'get_tzdb'
    auto& db = get_tzdb();
               ^
1 error generated.
````

**Comment by HowardHinnant** [comment]
This program segfaults:
````
#include "tz.h"
#include <iostream>

int
main()
{
    using namespace std::chrono;
    using namespace date;
    std::cout << make_zoned(std::string("Pacific/Wallis"), system_clock::now()) << '\n';
}
````
Stack trace:
````
Exception Type:        EXC_BAD_ACCESS (SIGSEGV)
Exception Codes:       KERN_INVALID_ADDRESS at 0x0000000000000000
Exception Note:        EXC_CORPSE_NOTIFY

Termination Signal:    Segmentation fault: 11
Termination Reason:    Namespace SIGNAL, Code 0xb
Terminating Process:   exc handler [0]

VM Regions Near 0:
--> 
    __TEXT                 000000010955b000-0000000109592000 [  220K] r-x/rwx SM=COW  /Users/USER/*/*.out

Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   a.out                         	0x00000001095845d7 date::tzfile_zone::get_info_impl(std::__1::chrono::time_point<std::__1::chrono::system_clock, std::__1::chrono::duration<long long, std::__1::ratio<1l, 1l> > >, int) const::$_0::operator()(std::__1::chrono::time_point<std::__1::chrono::system_clock, std::__1::chrono::duration<long long, std::__1::ratio<1l, 1l> > > const&, std::__1::__wrap_iter<date::detail::transition const*> const&) const + 119 (tz.cpp:3251)
1   a.out                         	0x000000010958454d bool std::__1::__invoke_void_return_wrapper<bool>::__call<date::tzfile_zone::get_info_impl(std::__1::chrono::time_point<std::__1::chrono::system_clock, std::__1::chrono::duration<long long, std::__1::ratio<1l, 1l> > >, int) const::$_0&, std::__1::chrono::time_point<std::__1::chrono::system_clock, std::__1::chrono::duration<long long, std::__1::ratio<1l, 1l> > > const&, std::__1::__wrap_iter<date::detail::transition const*> const&>(date::tzfile_zone::get_info_impl(std::__1::chrono::time_point<std::__1::chrono::system_clock, std::__1::chrono::duration<long long, std::__1::ratio<1l, 1l> > >, int) const::$_0&&&, std::__1::chrono::time_point<std::__1::chrono::system_clock, std::__1::chrono::duration<long long, std::__1::ratio<1l, 1l> > > const&&&, std::__1::__wrap_iter<date::detail::transition const*> const&&&) + 109 (__functional_base:437)
2   a.out                         	0x00000001095843c9 std::__1::__function::__func<date::tzfile_zone::get_info_impl(std::__1::chrono::time_point<std::__1::chrono::system_clock, std::__1::chrono::duration<long long, std::__1::ratio<1l, 1l> > >, int) const::$_0, std::__1::allocator<date::tzfile_zone::get_info_impl(std::__1::chrono::time_point<std::__1::chrono::system_clock, std::__1::chrono::duration<long long, std::__1::ratio<1l, 1l> > >, int) const::$_0>, bool (std::__1::chrono::time_point<std::__1::chrono::system_clock, std::__1::chrono::duration<long long, std::__1::ratio<1l, 1l> > > const&, std::__1::__wrap_iter<date::detail::transition const*> const&)>::operator()(std::__1::chrono::time_point<std::__1::chrono::system_clock, std::__1::chrono::duration<long long, std::__1::ratio<1l, 1l> > > const&, std::__1::__wrap_iter<date::detail::transition const*> const&) + 73 (functional:1437)
3   a.out                         	0x0000000109576215 std::__1::function<bool (std::__1::chrono::time_point<std::__1::chrono::system_clock, std::__1::chrono::duration<long long, std::__1::ratio<1l, 1l> > > const&, std::__1::__wrap_iter<date::detail::transition const*> const&)>::operator()(std::__1::chrono::time_point<std::__1::chrono::system_clock, std::__1::chrono::duration<long long, std::__1::ratio<1l, 1l> > > const&, std::__1::__wrap_iter<date::detail::transition const*> const&) const + 181 (functional:1817)
4   a.out                         	0x0000000109573327 date::tzfile_zone::get_info_impl(std::__1::chrono::time_point<std::__1::chrono::system_clock, std::__1::chrono::duration<long long, std::__1::ratio<1l, 1l> > >, int) const + 2903 (tz.cpp:3271)
5   a.out                         	0x00000001095727be date::tzfile_zone::get_info_impl(std::__1::chrono::time_point<std::__1::chrono::system_clock, std::__1::chrono::duration<long long, std::__1::ratio<1l, 1l> > >) const + 46 (tz.cpp:3209)
6   a.out                         	0x000000010955cb84 date::sys_info date::tzfile_zone::get_info<std::__1::chrono::duration<long long, std::__1::ratio<1l, 1000000l> > >(std::__1::chrono::time_point<std::__1::chrono::system_clock, std::__1::chrono::duration<long long, std::__1::ratio<1l, 1000000l> > >) const + 68 (tz.h:582)
7   a.out                         	0x000000010955c74e date::basic_zoned_time<std::__1::chrono::duration<long long, std::__1::ratio<1l, 1000000l> >, date::tzfile_zone>::get_info() const + 46 (tz.h:1404)
8   a.out                         	0x000000010955c6a3 void date::to_stream<char, std::__1::char_traits<char>, std::__1::chrono::duration<long long, std::__1::ratio<1l, 1000000l> >, date::tzfile_zone>(std::__1::basic_ostream<char, std::__1::char_traits<char> >&, char const*, date::basic_zoned_time<std::__1::chrono::duration<long long, std::__1::ratio<1l, 1000000l> >, date::tzfile_zone> const&) + 35 (tz.h:1503)
9   a.out                         	0x000000010955c447 std::__1::basic_ostream<char, std::__1::char_traits<char> >& date::operator<<<char, std::__1::char_traits<char>, std::__1::chrono::duration<long long, std::__1::ratio<1l, 1000000l> >, date::tzfile_zone>(std::__1::basic_ostream<char, std::__1::char_traits<char> >&, date::basic_zoned_time<std::__1::chrono::duration<long long, std::__1::ratio<1l, 1000000l> >, date::tzfile_zone> const&) + 39 (tz.h:412)
10  a.out                         	0x000000010955c38e main + 222 (test.cpp:9)
11  libdyld.dylib                 	0x00007fffb47ae235 start + 1
````

**Comment by Erroneous1** [comment]
For the leap seconds I'm just going to surround with a try catch. In CMake I'm going to print out a warning if it can't find TZDIR/right/UTC. I'm also adding some CMake routines for finding the right zoneinfo directory.

I'm combining the tests you provided and adding them as tests/tz_test/make_zone.pass.cpp. I'm still working through some of the issues. Fixing the undefined reference to `tzrule_db::current_zone()` was easy enough. I also had to fix some issues with `make_zoned` where you couldn't specify a specific time zone without also explicitly specifying the duration.

It looks like I need to use `std::enable_if` on `basic_zoned_time` to make sure `class TimeZone` is a time_zone class and not something like `char`. I'll try to push a commit it in the next day or so.

**Comment by HowardHinnant** [comment]
Lack of leap second support has to be detected at compile-time via platform identification.  Subsequently use of the leap-second-aware clocks on that platform will fail at compile-time.  It would be very error prone for these clocks to silently work with no leap second list.

Thanks for adding to the tests.  The tz side of the library is not tested well.

Changes to the existing API will have to be accompanied by a large benefit.  For example, I am very unlikely to rename `time_zone` to `TimeZone`.  This would disrupt existing clients.  And the current name was chosen to mimic the naming convention in the std::lib, as this library is being proposed for standardization.

**Comment by Erroneous1** [comment]
I will add macros to remove the leaps data member from tzfile_zone if compiling on OSX and it hasn't explicitly been enabled.

Sorry, `TimeZone` is a template parameter for `basic_zoned_time`. `time_zone` is a typedef to either `tzrule_zone` or `tzfile_zone` (determined by TIMEZONE_DEFAULT macro). My goal is to be backwards compatible with the current API.

**Comment by HowardHinnant** [comment]
Ah, excellent.  My review so far is very shallow, relegated to evenings when I'm already tired. :-)

**Comment by Erroneous1** [comment]
I wonder if we could use [time2posix](https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man3/time2posix.3.html#//apple_ref/doc/man/3/time2posix) on OS X.

**Comment by HowardHinnant** [comment]
I just ran this program:
````
#include "date.h"
#include <iostream>
#include <time.h>

int
main()
{
    using namespace std;
    using namespace std::chrono;
    using namespace date;
    auto tt = time(nullptr);
    auto now = system_clock::now();
    cout << "tt  = " << tt << '\n';
    cout << "now = " << floor<seconds>(now).time_since_epoch() << '\n';
    cout << "t2p = " << time2posix(tt) << '\n';
    cout << "p2t = " << posix2time(tt) << '\n';
}
````
The output was:
````
tt  = 1494548350
now = 1494548350s
t2p = 1494548350
p2t = 1494548350
````

**Comment by Erroneous1** [comment]
ok so time2posix is a bust. I'm now falling back to defaulting to `tzrule_db` on OS X since `tzfile_db` isn't 100%. I'm hoping this commit fixes the issues you had before. I tested on gcc and clang on Linux.

File: tz.h [hunk 2]
```diff
@@ -51,7 +51,13 @@
...
 #  endif
 #  if !defined(TIMEZONE_FILES) && !defined(TIMEZONE_RULES)
-#    define TIMEZONE_FILES 1
+#    if !defined(__APPLE__)
+#      define TIMEZONE_RULES 1
+#      define TIMEZONE_FILES 0
+#    else
+#      define TIMEZONE_RULES 0
+#      define TIMEZONE_FILES 1
+#    endif
```
HowardHinnant: Looks like you got this exactly right, but backward. :-)

Erroneous1: Quite right! That's what I get for working on it right before bed. I ammended my commit.

HowardHinnant:
Why are there two separate types:  `tzfile_zone` and `tzrule_zone`?  Would it not be simpler to just have one type `time_zone`, with two different implementations?  That would make it easier to ensure that the public API is consistent, and eliminate the need for the changes to `make_zoned`.

HowardHinnant:
I'm not convinced that I want to support `locate_zone(zone_name, database_path)`.  Do you have a motivating use case for this API?

HowardHinnant:
Confirmed that `make_zoned("Pacific/Wallis", system_clock::now())` works now on macOS with `TIMEZONE_RULES=1`.

Erroneous1:
There are a few reasons why I broke it out into 2 (or more, potentially) time_zone and TZ_DB classes.

- There might be times when you want features from both. For instance, tzfile_zone loads much faster than tzrule_zone (and might operate faster, but haven't tested), but on OS X (and potentially some Linux/BSD variants) won't have leap second calculations. Software that wanted to be quick and have leapseconds might want tzfile_zone with tzrule_db's leap seconds.
- We can do extra testing like comparing one time_zone class's results against another without saving output to a file can comparing the output.
- By opening up `zoned_time` to `basic_zoned_time` we allow users to provide their own time_zone-like class.

I'm interested in your thoughts on the matter.

For locate_zone accepting a database path, it might be useful to be able to specify at run-time where to load tzfiles from. For instance, you can run zic on the latest IANA data and save to `~/.cache/zoneinfo` to get leapseconds and up to date tzfiles in OS X. On Windows we could probably do something similar, though I probably want to add code to replace `'/'` with `folder_delimiter`. We could also do something in the future like use tzrule_db to generate tzfiles and then use tzfile_db (though zic does this). Another option of course is for tzfile_db to contain the tzdir that can be setup during `init_tzdb` so that it only has to set done once.

HowardHinnant:
I'm confused about the purpose of `TIMEZONE_DEFAULT`.  It appears to be equivalent to `!TIMEZONE_FILES` and also equivalent to `TIMEZONE_RULES`.  I ask because currently when I `#include "tz.h"` I get:
````
../date/tz.h:122:4: error: "Cannot use TIMEZONE_MAPPING as TIMEZONE_DEFAULT if TIMEZONE_MAPPING is disabled!"
#  error "Cannot use TIMEZONE_MAPPING as TIMEZONE_DEFAULT if TIMEZONE_MAPPING is disabled!"
   ^
````
I suspect it would be simpler to ditch `TIMEZONE_DEFAULT` and just use `TIMEZONE_RULES` and `TIMEZONE_FILES`. 

Also I suspect you are misinterpreting `TIMEZONE_MAPPING`.  This is a Windows-only facility that maps Windows timezone names to IANA timezone names.  It isn't used on macOS or Linux.  And it is only needed for the Windows implementation of `current_zone()` which discovers the current Windows timezone and has to translate that to a IANA timezone.

On Windows, we will always have to use `TIMEZONE_RULES` as the compiled IANA timezone files will never be found on Windows.  `TIMEZONE_FILES` will work on macOS (`__APPLE__`), but we will need to `#ifdef` out code for `utc_clock`, `tai_clock`, and `gps_clock` in `tz.h`.  That is, wrap all of that code in `#if !DATE_TIMEZONE_FILES_NO_LEAP`.

**Comment by HowardHinnant** [comment]
I see, thanks.  I will give this more thought.

**Comment by Erroneous1** [comment]
I'll have to clean that up. Originally I was going to split the Windows
timezone mapping into a 3rd time zone class but after getting into the
details I found that it didn't make sense to do so. TIMEZONE_DEFAULT is
used to override which zone and db classes are typedef'd as time_zone and
TZ_DB, which are used in things like make_zoned and for leap seconds.

I tested using cmake on OS X and running ctest and had no issues after my
last patch. I have another small patch for Windows I'm working on, but most
of the tests won't compile on VS 2013 due to lack of constexpr for
static_assert. Validate ran though. I'll try installing a community version
of VS to test.

Btw are you aware of date.parse.pass.cpp failing an assert under gcc but
not on clang? I can post the exact line tomorrow.

**Comment by HowardHinnant** [comment]
> Btw are you aware of date.parse.pass.cpp failing an assert under gcc but
not on clang?

I'm answering without looking into it (going to bed now).  But it is a policy of mine to let tests (but not library code) fail when compilers have a bug.  This may be a bug related to the use of `{}` as opposed to `()` in gcc-6.1 and earlier.  That one has bitten a lot.

**Comment by Erroneous1** [comment]
Regarding the error, I'm running gcc 6.3.1.

```
date_test_parse_pass: ../test/date_test/parse.pass.cpp:48: void test_a(): Assertion `!in.fail()' failed.
```

It occurs here:
```c++
        // correct abbreviation
        std::istringstream in{"Sun 2016-12-11"};
        sys_days tp;
        in >> parse("%A %F", tp);
        assert(!in.fail()); //fail
        assert(!in.bad());
        assert(!in.eof());
        assert(tp == 2016_y/12/11);
```

Which is odd. It doesn't occur on the section above it which uses `in >> parse("%a %F", tp);`. From what I could see, `%a` and `%A` should be equivalent.

I managed to get clang to produce the error by using `-std=gnu++14 -I/usr/include/c++/6.3.1` and not using `-stdlib`.

HowardHinnant:
This may be a bug wherein gcc doesn't parse weekday names in a case-insensitive manner.  I know this is a gcc bug, but I'm not 100% positive that this bug is the cause of this test failure.  VS has a similar bug.  Both std::lib authors have been notified of it.

Erroneous1:
Ok, sounds good. I'm not worried about it, but I figured that not many people run the tests other than you. Maybe having cmake support will change that since it is so easy to do with `mkdir build && cd build && cmake .. && make && ctest`.

I'll look into moving the native zone mapping to `detail::current_zone_string` and adding `locate_native_zone` to tzfile_zone.

HowardHinnant:
I just learned I can no longer say `get_tzdb()`.  The API direction this is taking isn't going to work for me.  The overall goal:  Have an option to use the compiled tz files, I support.  And I really appreciate the work you've done on that front, and plan to merge it.  But the changes to the API I can't support.

I'm going to work on a variant of your PR which takes your core logic, and integrates it into what is currently on the master branch with a minimum of changes, and with a single API for each major class (e.g. `time_zone` and `get_tzdb()`).  I don't plan on supporting the class-nested API you've added, nor the ability to run both databases at once.

All additions or changes to the API will have to be documented in https://github.com/HowardHinnant/date/blob/gh-pages/tz.html and my current goal is to only have to update the installation instructions with a couple of new macros, and restrictions on the use of the OS-supplied tz files.

One thing I haven't seen yet is a copyright from you, especially in tz.h and tz.cpp.  This is not required, but is encouraged.

The CMake work should be a different PR.

Erroneous1:
If you can't call `get_tzdb()` then that was an ommission on my part and it should be an inline function:
`inline const TZ_DB& get_tzdb() { return TZ_DB::get_tzdb(); }`. Similar inline functions should be defined for `current_zone()` and `locate_zone(const std::string&)`. My goal was to require no changes to the current API, but to allow for expanded functionality. Are you planning on keeping `basic_zoned_time` so that a user could still provide their own time_zone-like class?

I thought I added my name to tz.h but I can review it tonight. I just realized I never put my name in my Github bio, which is probably what you were looking for. I should probably also add license information to the tests I added and CMakeLists.txt

If you'd like me to refactor it as a single class I can.

I will make a separate PR for CMake.

HowardHinnant:
My current plan (and it is just a plan at this point) is to try to make just one class called `time_zone`, that is `#ifdef`d on the database type (files/rules).  I won't know for sure if that is the way to go until I try it.

But regardless of how it ends up, you've made very significant contributions here and I would much prefer to have your copyright in the files.

HowardHinnant:
👍

tz.cpp
```diff
@@ -3087,121 +3124,145 @@ locate_native_zone(const std::string& native_tz_name)
...
+    size_t buffer_size = time_size + 4;
+    for(std::uint32_t leap_last = 0; leap_last < leap_count;) {
+        if(!in.read(buffer.data(), buffer_size))break;
+        std::uint32_t count = detail::interpret32(&buffer[time_size]);
+        if(++leap_last != count) {
+            throw std::runtime_error{"invalid tzfile"};
+        }
+        leaps_.emplace_back(sys_seconds{std::chrono::seconds{detail::interpret_time(buffer.data(), time_size)}});
+    }
+    if(!leaps_.empty()) {
```
HowardHinnant: Is `!leaps_.empty()` ever true when reading files not under `right`?  I'm wondering if this part can just be ignored.

Erroneous1:
Leaps are empty for all files in the posix sub-tree, but certain
implementations may prefer to use right instead of posix. If a file has
leaps then the all times listed will be number of seconds since epoch
including leaps so they have to be adjusted.

HowardHinnant:
Update.  This is still under review.  There's been a lot of progress, but not ready to push yet.  Just wanted to let you know this is not being dropped.

Erroneous1:
Sounds good. I made another PR for CMake a while ago so you can separate it out. When you get a final commit for TZ files I'll probably need to update the CMake PR.

BTW you may want to use the `set_install()` function instead of passing the tzdir around. That would let you have the full functionality from passing the tzdir without having to do it all of the time. I don't think we need to worry about thread safety much since it should just be set once on init.

I would probably keep a TZDIR macro so that it can be configured at compile time. CMake can figure out where the TZDIR is in the system pretty easily. I'm not sure how you want to handle leap seconds for systems that don't have the `right/` variant tzfiles, but CMake can detect at compile time whether or not `${TZDIR}/right/UTC` exists so it'd be easy to disable.

HowardHinnant:
My thought process to-date has been that if you're using the binary files, it is because they are supplied by your OS.  This is because they contain a strict subset of the information of the text files.  With `DHAS_REMOTE_API=0` and `-O3`, the performance advantage of the binary files becomes relatively insignificant.

So INSTALL becomes wherever the OS has installed the binary files.  It is no longer configurable.

On Linux, the binary files have the highest quality.   On Apple, they exist, but kind of suck.  Still one might prefer them.  They have limited range (1904-2038), and lack leap second information.  On Windows they don't exist at all, and the text DB is your only option.  I suspect that the most popular use of the binary files will be on mobile (Android/iOS) where installing the text DB is most problematic.

HowardHinnant:
Ready for review: https://github.com/HowardHinnant/date/tree/tzfile

HowardHinnant:
This is now merged on to master: https://github.com/HowardHinnant/date/commit/a610f087c1064896f53ac3d4b2ef181f740d1d7f

PR #155, opened 2017-05-16
Implemented cmake

Erroneous1:
Reconfigured .fail tests to use static_assert to work with ctest
Reconfigured testit to work with .fail tests
Can be used by calling add_subdirectory(date)
Can turn on/off testing and various tz options
Added a couple tests for time_zones


CMakeLists.txt
```diff
@@ -0,0 +1,160 @@
...
+		option(DATE_TZ_NO_DATE_TZ_SHELL  "Disable using shell programs to manage IANA Time Zone rules" Off)
+	else()
+		option(DATE_TZ_NO_DATE_TZ_SHELL  "Disable using shell programs to manage IANA Time Zone rules" On)
+	endif()
+endif()
+
+message(STATUS "Using ${CMAKE_CXX_COMPILER} (compiler id: ${CMAKE_CXX_COMPILER_ID})")
+if ("${CMAKE_CXX_COMPILER_ID}" MATCHES ".*Clang")
+	if(NOT "${CMAKE_CXX_FLAGS}" MATCHES "-stdlib")
+		set(CMAKE_CXX_FLAGS "-std=c++14 -stdlib=libc++ ${CMAKE_CXX_FLAGS}")
```
remyabel: This assumes that the user has libc++, which is only true if they built it from source on most systems.

CMakeLists.txt
```diff
@@ -0,0 +1,160 @@
...
+		add_definitions(-DHAS_REMOTE_API=0)
+		add_definitions(-DAUTO_DOWNLOAD=0)
+		set(TZLINKS "")
+	endif()
+
+	add_library(date-tz STATIC tz.cpp)
+
+	target_link_libraries(date-tz ${TZLINKS})
+
+	set_property(TARGET date-tz PROPERTY CXX_STADARD 14)
```
remyabel: CXX_STADARD is misspelled.

CMakeLists.txt
```diff
@@ -0,0 +1,160 @@
...
+		
+		add_test(validate validate)
+		if(NOT MSVC)
+			set(DATE_TZ_DO_FAIL_PASS On)
+		elseif(${MSVC_VERSION} GREATER 1910)
+			set(DATE_TZ_DO_FAIL_PASS On)
+		endif()
+		if(DATE_TZ_DO_FAIL_PASS)
+			file(GLOB_RECURSE DATE_STATIC_TESTS RELATIVE ${CMAKE_SOURCE_DIR}/test/ *.fail.cpp)
+			file(GLOB_RECURSE DATE_RUNTIME_TESTS RELATIVE ${CMAKE_SOURCE_DIR}/test/ *.pass.cpp)
```
remyabel: GLOB_RECURSE is considered bad practice because it requires invoking cmake again to update the filelist, which can be problematic with version control.

CMakeLists.txt
```diff
@@ -0,0 +1,160 @@
...
+			foreach(source_file ${DATE_STATIC_TESTS} ${DATE_RUNTIME_TESTS})
+				string(REPLACE "." "_" source_target ${source_file})
+				string(REPLACE "/" "_" source_target ${source_target})
+				string(REPLACE "\\" "_" source_target ${source_target})
+				string(REPLACE "_cpp" "" source_target ${source_target})
+				add_executable(${source_target} ${CMAKE_SOURCE_DIR}/test/${source_file})
```
remyabel: Why not just use a unit test library?

CMakeLists.txt
remyabel: [Generator expressions](https://cmake.org/cmake/help/v3.2/manual/cmake-generator-expressions.7.html) combined with `target_compile_options` could prove useful here.

Erroneous1:
Thanks for all of the suggestions. Time has gotten away from me so I won't have a chance to work on this. I added you as a collaborator so that you can contribute to it.

daixtrose:
Oops! I overlooked this PR and [implemented another proposal](https://github.com/HowardHinnant/date/pull/210#issuecomment-330665223)

Maybe the ideas could be merged in some way or the other. I concentrated on packaging and running the tests (though not yet happy with my approach)

CMakeLists.txt
```diff
@@ -0,0 +1,189 @@
...
+    set(NEED_PARENT On)
+    set(DATE_DEFAULT_ENABLE_TESTS Off)
+endif()
+
+find_package(CURL)
+
+if(WIN32)
+    find_program(DATE_TZ_SHELL_REMOVE_DIRECTORY "rd")
+    find_program(DATE_TZ_SHELL_REMOVE_FILE "del")
+    find_program(DATE_TZ_SHELL_MOVE_FILE "move")
```
daixtrose: What about `${CMAKE_COMMAND} -E rename ...`. In general: are [cmake file operations](https://cmake.org/cmake/help/latest/command/file.html) inappropriate here?

CMakeLists.txt
```diff
@@ -0,0 +1,189 @@
...
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+cmake_minimum_required(VERSION 3.2 FATAL_ERROR)
+project(date VERSION 2.1 LANGUAGES CXX)
+
+set(_CMAKE_TOOLCHAIN_PREFIX llvm-)
```
daixtrose: Why ist llvm- defined at uch a prominent position. Shouldn't the system work with any compiler?

CMakeLists.txt
```diff
@@ -0,0 +1,189 @@
...
+
+if(${CMAKE_SOURCE_DIR} STREQUAL ${CMAKE_CURRENT_SOURCE_DIR})
+    set(NEED_PARENT Off)
+    set(DATE_DEFAULT_ENABLE_TESTS On)
+else()
+    set(NEED_PARENT On)
+    set(DATE_DEFAULT_ENABLE_TESTS Off)
+endif()
+
+find_package(CURL)
```
daixtrose: file(DOWNLOAD <url> <file> [<options>...]) ?

CMakeLists.txt
remyabel: Given that it's an internal variable it should probably be removed altogether.

remyabel:
I've kind of lost interest in this. Feel free to integrate the two PR's, but I cannot close this since I'm not the author.

Erroneous1:
Since I cannot devote time to this right now I'm closing the PR. Feel free to use this as a skeleton in your own PR

PR #167, opened 2017-06-29
Also ignore leap-seconds.list

RockinRoel:
The tzdata package on Ubuntu 16.04, maybe other distributions too, includes /usr/share/zoneinfo/leap-
seconds.list. This is not a TZif file, so it should be ignored.


HowardHinnant:
Thanks!

PR #188, opened 2017-07-27
Add default constructor for weekday_indexed.  Fixes #187

### LINKED ISSUE #187: Default constructor of date::year_month_weekday makes call to implicitly deleted constructor
The code

```date::year_month_weekday ymwd;```

does not compile because the `date::weekday_indexed` member of `date::year_month_weekday` declares constructors, but not a default one, so the default constructor is implicitly deleted, thus `date::year_month_weekday` is not default constructable.

pauluap:


HowardHinnant:
My apologies.  For some reason I was in too much of a rush and didn't see this P/R.  I would much rather your name be on the commit than mine.  I want to encourage participation.  Perhaps in the future, on the issue indicate that you have, or are in process of submitting a P/R.  Yours will definitely take precedence over mine.

pauluap:
No worries, you made an additional change that I didn't anyway.  I think that next time I have a solution ready, I'll submit it as a stand-alone PR instead of opening an issue then a PR.

HowardHinnant:
That works too!  Much thanks for your help.

PR #189, opened 2017-08-12
Fix compilation warnings for -DUSE_OS_TZDB.

jankratochvil:
tz.cpp:1617:24: error: unused parameter 't' [-Werror,-Wunused-parameter]
maybe_reverse_bytes(T& t, std::false_type)
                       ^
tz.cpp:2470:1: error: unused function 'get_version' [-Werror,-Wunused-function]
get_version()
^
tz.cpp: In lambda function:
tz.cpp:1922:84: error: declaration of ‘t’ shadows a previous local [-Werror=shadow]
                                        [](const sys_seconds& x, const transition& t)
                                                                                    ^
tz.cpp:1921:19: note: shadowed declaration is here
         for (auto t = std::upper_bound(transitions_.begin(), transitions_.end(), l,
                   ^

### DIFF
--- a/tz.cpp
+++ b/tz.cpp
@@ -1614,7 +1614,7 @@ template <class T>
 static
 inline
 void
-maybe_reverse_bytes(T& t, std::false_type)
+maybe_reverse_bytes(T&, std::false_type)
 {
 }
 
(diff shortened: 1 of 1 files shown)


HowardHinnant:
Thanks!

PR #195, opened 2017-08-31
Build vs2015

kirylll:
Updated code to build with vs 2015 and rmt_curl nuget package

### DIFF
--- a/tz.cpp
+++ b/tz.cpp
@@ -142,7 +142,7 @@
 #if HAS_REMOTE_API
    // Note curl includes windows.h so we must include curl AFTER definitions of things
    // that effect windows.h such as NOMINMAX.
-#if defined(_MSC_BUILD) && defined(SHORTENED_CURL_INCLUDE)
+#if defined(_MSC_VER) && defined(SHORTENED_CURL_INCLUDE)
    // For rmt_curl nuget package
 #  include <curl.h>
 #else
--- a/tz.cpp
+++ b/tz.cpp
@@ -142,7 +142,12 @@
 #if HAS_REMOTE_API
    // Note curl includes windows.h so we must include curl AFTER definitions of things
    // that effect windows.h such as NOMINMAX.
+#if defined(_MSC_BUILD) && defined(SHORTENED_CURL_INCLUDE)
+   // For rmt_curl nuget package
+#  include <curl.h>
+#else
 #  include <curl/curl.h>
+#endif
 #endif
 
 #ifdef _WIN32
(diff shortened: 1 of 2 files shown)


HowardHinnant:
Thanks much for your efforts to keep us running on VS 2015!

I've just committed a test on master that shows breakage with this commit: https://github.com/HowardHinnant/date/commit/aad60108313e4ebe9db3b4cdc3df9e4de3765fba
```
Jade:~/Development/date/test/tz_test> testit
In file included from OffsetZone.pass.cpp:25:
/Users/howardhinnant/Development/date/tz.h:425:61: error: no matching member function for call to 'to_sys'
                      decltype(std::declval<TimeZonePtr>()->to_sys(local_time<Duration>{},
                               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~
OffsetZone.pass.cpp:79:52: note: in instantiation of template class 'date::zoned_time<std::__1::chrono::duration<long long, std::__1::ratio<1,
      1000000> >, OffsetZone>' requested here
    zoned_time<system_clock::duration, OffsetZone> zt{OffsetZone{offset}, now};
                                                   ^
OffsetZone.pass.cpp:48:9: note: candidate function template not viable: requires single argument 'tp', but 2 arguments were provided
        to_sys(date::local_time<Duration> tp) const
        ^
1 error generated.
/Users/howardhinnant/Development/date/test/tz_test/OffsetZone.pass.cpp failed to compile
Compile line was: clang++ -std=c++14 -I/Users/howardhinnant/Development/date -Wall /Users/howardhinnant/Development/date/tz.cpp -lcurl OffsetZone.pass.cpp
```
I suspect that the way to get VS-2015 building is to add #ifdef's like this around the entire template declaration:  https://github.com/HowardHinnant/date/blob/master/tz.h#L362-L365 adjusting the `_MSC_VER` as necessary for VS-2015 (I don't know which version 1911 refers to).

HowardHinnant:
Ping!  I'd love for this library to work on VS 2015, but I need help from people with VS 2015 installed to do so.  This particular patch breaks things, and I added a test to help demonstrate that.  But I believe there are other things we could do such as removing the entire template constraint for VS 2015.  I just need a VS-2015-enabled volunteer to experiment with that approach.

kirylll:
Hi Howard, Sorry for my late reply.
I’ve checked that these template constraints doesn’t work properly with latest vs 2017 as well.
So I committed an update where these constraints are disabled for all vs version by adding #ifdef  _MSC_VER.
1911 is the latest version now available of vs 2017.

HowardHinnant:
Merged: https://github.com/HowardHinnant/date/commit/41563c46e870accd7adc97b9627f5f00e9dcd6b2

You'll have to `git reset --hard` to my master as I squashed and reordered your commits.

Thanks much!

PR #202, opened 2017-09-08
[FreeBSD] fix #201

### LINKED ISSUE #201: Fails to get current timezone on FreeBSD
image

image

The library experts `/etc/localtime` to be a symbolic link. However on FreeBSD it is not. If you look at the image above it is an actual `tzfile`. To get the timezone in this case we might have to parse the file. Here is my output to zdump

image

atkawa7:


HowardHinnant:
Merged: https://github.com/HowardHinnant/date/commit/5f01382e248d77e1c207620e400096ce33837159

Thanks!

PR #203, opened 2017-09-09
Fix CHAR_BIT not found on GCC 6.3.0

neuraleskimo:
Building with GCC 6.3.0 fails on date.h line 1046 because the CHAR_BIT macros isn't found. It appears that limits.h is included in the headers of some implementations, but not GCC 6.3.0. The fix is simply include climits (of limits.h if you prefer).

### DIFF
--- a/date.h
+++ b/date.h
@@ -40,6 +40,7 @@
 #include <cstdint>
 #include <cstdlib>
 #include <ctime>
+#include <climits>
 #include <ios>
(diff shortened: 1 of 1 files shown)


HowardHinnant:
Merged: https://github.com/HowardHinnant/date/commit/ef6d53595de9bda548c4d705f40ab6bc8eaba1e9

Thanks!

PR #210, opened 2017-09-19
Proposal for an alternative directory structure

daixtrose:
With a project structure
```
<project_name>
|-include
   |-<project_name>
     |- <header>.hpp
```
IMHO it is much easier to integrate a project into other projects. I prefer 
```c++
#include "date/date.hpp"
```
over
```c++
#include "date.hpp"
```
Though the files could be easily copied around locally by myself, I nonetheless request to reorganize the file structure in order to allow easy integration of this library e.g. via git submodules (which allows me to always include the latest and greatest version). See e.g. @ericniebler´s [include/range/v3](https://github.com/ericniebler/range-v3/tree/master/include/range/v3) which follows my proposed strategy (AFAIK this is a boost library convention).


galik:
I like this idea. Having the headers in a 'project-named' subdirectory makes it less likely to conflict with other headers and should help reduce noise in places like `/usr/include`.

daixtrose:
I am alsi trying to create a CMakeLists.txt for easy import and packaging. See https://github.com/daixtrose/continuous-integration-demo for where I start from.

daixtrose:
I just added a first shot for CMake packaging. 
- new file:   CMakeCPack.cmake
- new file:   CMakeCPackConfig.cmake.in
- new file:   CMakeLists.txt

i also fixed the path to headers in 
- modified:   src/tz.cpp

With these changes it is possible to easily create a distributable package by issuing the commands
```
mkdir build
cd build
cmake ..
make 
make package
```

daixtrose:
@HowardHinnant I also added CMake files for the tests. Now you can do

```
mkdir build
cd build
cmake ..
make 
make test
make package
``` 

There were some compilation errors which I commented out in *.pass.cpp and marked them with a "TODO:" so you can easily find them

HowardHinnant:
Thanks for your work on this.  Not ignoring, just busy at my day job.

CMakeCPack.cmake
```diff
@@ -0,0 +1,36 @@
...
+LIST(APPEND CPACK_SOURCE_IGNORE_FILES "Makefile")
+LIST(APPEND CPACK_SOURCE_IGNORE_FILES "Makefile\\\\..*")
+LIST(APPEND CPACK_SOURCE_IGNORE_FILES "CMakeCache.txt")
+LIST(APPEND CPACK_SOURCE_IGNORE_FILES "CMakeFiles")
+LIST(APPEND CPACK_SOURCE_IGNORE_FILES "cmake_install.cmake")
+LIST(APPEND CPACK_SOURCE_IGNORE_FILES "install_manifest.txt")
+LIST(APPEND CPACK_SOURCE_IGNORE_FILES "CPackConfig.cmake")
+LIST(APPEND CPACK_SOURCE_IGNORE_FILES "CPackSourceConfig.cmake")
+
+set(CPACK_GENERATOR "DEB" "TGZ" "ZIP")
```
remyabel: Can we add a RPM generator as well for non-Debian systems?

test/CMakeLists.txt
```diff
@@ -0,0 +1,68 @@
...
+
+################################################################################
+# Import Catch Test Framework 
+################################################################################
+
+# See http://dbgit.prakinf.tu-ilmenau.de/code/pfabric/blob/dbe8dbfe21a79c00738cf7762368181edf249767/cmake/Download3rdParty.cmake
+
+include(${CMAKE_SOURCE_DIR}/cmake-helper/DownloadProject/DownloadProject.cmake)
+
+download_project(PROJ               catch
```
remyabel: As much as I like Catch, I don't see how it can be integrated here without rewriting all the tests.

CMakeLists.txt
```diff
@@ -0,0 +1,108 @@
...
+# Put config file in per-project dir (name MUST match), can also
+# just go into 'cmake'.
+install(EXPORT ${LIBRARY_NAME} DESTINATION share/${LIBRARY_NAME}/cmake)
+
+# This makes the project importable from the build directory
+export(TARGETS ${LIBRARY_NAME} FILE ${LIBRARY_NAME}.cmake)
+
+###############################################################################
+# Packaging configuration
+if (${LIBRARY_NAME}_STANDALONE_PROJECT)
```
remyabel: Again this seems Debian specific and RPATH sets off a red flag for me (though maybe that's a problem only on Debian). Surely there's a better way to handle other distros as well?

CMakeCPack.cmake
daixtrose: I believe that simply adding "RPM" in line 30 is sufficient. I just did not have an rpm system at hand. Feel free to add it, test it and commit or PR into this branch.

@HowardHinnant FYI: the installer is still lacking the consideration of the IANA stuff, but i would like to address this in a separate branch after merge.

CMakeLists.txt
daixtrose: CPack allows a per target configuration. What you see here is indeed the Debian specific part, which is ignored on other platforms. For RPM we have to add a similar section. On windows everything goes to the standard places (admittedly not tested yet).

With regard to RPATH: This configuration installs into `/opt/date-VERSION` and I tried all variants with and without RPATH in [continuous-integration-demo](https://github.com/daixtrose/continuous-integration-demo) and came to the conclusion that a relative RPATH search, i.e. an executable that prefers its **own** library is exactly what you need in case you have an executable.

I kept this entry here because I was thinking about an option to package some of the test executables into the distro. Not sure though ...

daixtrose:
@HowardHinnant Maybe it was not so clever to change the directory structure **and** try to add CMake support in the very same PR. This was only motivated by the trouble I had adding the libtrary to my own project. Both changes are necessary in the long term, but the directory structure itself is more important.

Please let me know whether we should go forward from here (fixing some Cmake issues, adding RPM, etc.) or if I should provide a PR containing only the path restructuring part, leaving the build and test system as it is.

test/CMakeLists.txt
daixtrose: IMHO not a big deal. The code is clearly structured. 

Tests not depending on the existence of `sh` is my aim here. I also just wanted to point out how easy it is to incorporate `Catch` with `CMake`, which opens the option for a more flexible CI.

HowardHinnant:
It looks like you have the directory structure change in just this one commit 1bb5409a4301d4394effe6a361bda204d77403a9 right?  I can consider and merge that one commit in isolation.

daixtrose:
Yes, but there might be some issues with `#include ...` statements and I forgot to fix the lines in the test script, but that is easy to add.

daixtrose:
So if you merge 1bb5409 I can rebase this branch and continue with the CMake proposal.

HowardHinnant:
Ok, master is updated to your new directory structure.

HowardHinnant:
I believe the full intent of this P/R is now on master.

daixtrose:
Agreed, and sorry for not providing a solution in time. I was stuck with an unresolved issue about propagating include directories. I will take a look at what is now in master and maybe my contribution is not necessary anymore.

HowardHinnant:
No worries!  You got the ball rolling!  Thanks!

PR #216, opened 2017-10-03
Fix uninitialized variable use in to_stream()

tomba:
tm variable is not initialized in to_stream(), and valgrind warns about
"Conditional jump or move depends on uninitialised value(s)".

Initialize tm to zero, and with that done, we can drop tm
initializations from two particular code paths.

### DIFF
--- a/include/date/date.h
+++ b/include/date/date.h
@@ -4566,7 +4566,7 @@ to_stream(std::basic_ostream<CharT, Traits>& os, const CharT* fmt,
 {
     using namespace std;
     using namespace std::chrono;
-    tm tm;
+    tm tm{};
 #if !ONLY_C_LOCALE
     auto& facet = use_facet<time_put<CharT>>(os.getloc());
 #endif


include/date/date.h
```diff
@@ -5323,7 +5322,6 @@ to_stream(std::basic_ostream<CharT, Traits>& os, const CharT* fmt,
                     os << CharT{'%'} << modified << *fmt;
                 else
                 {
-                    tm = std::tm{};
```
HowardHinnant: I'd like to keep the two zeroing places I'm already doing it.  Those are places where I need to clear the `tm` from the scribbling of prior commands into it.

tomba: Oh, sorry, I missed the fact that there's a loop, so tm gets reused.

In that case, wouldn't it be better to have the tm declared and initialized inside the for loop? Well, seems that man strftime() gives the exact fields each formatting uses, so with careful coding it saves a bit of time to not initialize all of it.

Hmm, or maybe that's not a safe way, as valgrind complained that someone was using uninitialized data, so clearly tm_wday was not the only field being used from tm.

HowardHinnant: My best guess is that the valgrind report is a false positive, and the single zeroing prior to the loop is a cheap way to silence that false positive.

tomba: I looked at libc's strftime. It reads tm_hour, but only uses it for the hour based formats. So yes, zeroing tm before the loop should be fine.

HowardHinnant: You can either push another commit, in which case I'll squash your two commits and publish as one commit by you, or you can squash it yourself and push.  Either way works for me.

PR #221, opened 2017-10-12
Replacing 0 with nullptr to avoid GCC warning

morinim:
All the calls to `std::time_get::get` had `0` as end-of-range iterator.

E.g.

```
auto& f = use_facet<time_get<CharT>>(is.getloc());
// ...
f.get(is, 0, is, err, &tm, command, fmt+1);
```

Using `nullptr` instead of `0` doesn't trigger the GCC 5.x warning:

> warning: zero as null pointer constant [-Wzero-as-null-pointer-constant]


HowardHinnant:
Thanks!

PR #232, opened 2017-11-11
express reverse_bytes in a way that is easy to optimize

Erroneous1:
Optimizes to single bswap instruction on gcc and clang for x86_64
[godbolt](https://godbolt.org/g/jzt92d)

Only has overloads for int32 and int64 but that's all it is used for. No noticeable impact on running validate.cpp with -O2.

### DIFF
--- a/src/tz.cpp
+++ b/src/tz.cpp
@@ -1691,17 +1691,32 @@ enum class endian
     big    = __ORDER_BIG_ENDIAN__
 };
 
-template <class T>
 static
 inline
-void
-reverse_bytes(T& t)
+std::uint32_t
+reverse_bytes(std::uint32_t i)
 {
-    unsigned char* bytes = static_cast<unsigned char*>(std::memmove(std::addressof(t),
-                                                                    std::addressof(t),
-                                                                    sizeof(T)));
-    for (unsigned i = 0; i < sizeof(T)/2; ++i)
-        std::swap(bytes[i], bytes[sizeof(T)-1-i]);
+    return
+        (i & 0xff000000u) >> 24 |
+        (i & 0x00ff0000u) >> 8 |
+        (i & 0x0000ff00u) << 8 |
+        (i & 0x000000ffu) << 24;
+}
+
+static
+inline
+std::uint64_t
+reverse_bytes(std::uint64_t i)
+{
+    return
+        (i & 0xff00000000000000ull) >> 56 |
+        (i & 0x00ff000000000000ull) >> 40 |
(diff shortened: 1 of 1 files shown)


wolfv:
Why not `int32_t __builtin_bswap32 (int32_t x)` & `int64_t __builtin_bswap64 (int64_t x)` for gcc/clang and for Visual Studio:

`intrin.h`

`unsigned __int64 _byteswap_uint64(unsigned __int64 value);`

all according to: https://stackoverflow.com/questions/105252/how-do-i-convert-between-big-endian-and-little-endian-values-in-c

Erroneous1:
We could do a number of `#ifdef`s to see which platform we're on, which header to include, and which bswap function to call (why not `#include <byteswap.h>` and use `__bswap_32`?). But with GCC and Clang, the 2 major compilers that will use this code since it only applies to using OS TzDB, this code will compile to a single instruction and also work on other compilers regardless of the headers available. If I can avoid including new headers I prefer to.

We could implement a complete approach like in boost intrinsic but in my limited testing there is already little performance difference at runtime and this patch covers the vast majority of use cases.

HowardHinnant:
Sorry for coming late to the party.  I've been traveling for the past two days.

I appreciate this issue being looked into.  The current code is already optimal for clang.  I just assumed gcc would optimize it too.  It would be nice to get it optimized on both clang and gcc with minimal fuss.

Erroneous1:
The [current implementation](https://godbolt.org/g/6h5KvG) does a `mov` for each byte for both GCC and Clang with -O3. Clang does this at -O2 and -O3 while GCC doesn't unroll the loop for int64_t and uses a `movzx` and a `mov` per byte on -O2.  I assume that a single `bswap` instruction is actually faster than 8 or 16 `mov` instructions, but I haven't tested that.

HowardHinnant:
Interesting.  I'm not yet sure what the (important) difference is: https://godbolt.org/g/VcTAcb

HowardHinnant:
I didn't realize it was that fragile.  The intrinsics are beginning to sound pretty attractive, maybe with a portable fall-back for easy porting to other unanticipated platforms.

HowardHinnant:
`<aside>` Beman presented his boost::endian lib at Albuquerque last week and it was favorably viewed.  Perhaps by C++20 we can stop worrying about this problem! :-)

HowardHinnant:
Merged: https://github.com/HowardHinnant/date/commit/c9ef0a8f05f1920572a7bee7c813f46c6c5e4eb9

Thanks!

PR #234, opened 2017-11-15
properly detect timezone on buildroot/uclibc systems

paresy:
On Buildroot with uclibc the timezone is a symlink in /etc/TZ which points to the zoneinfo which is located in /usr/share/zoneinfo/uclibc/*.

Have a look at the tz package from the buildroot sources. https://git.busybox.net/buildroot/tree/package/tz/tz.mk#n33

In line 33 the symlink is created which shows the source and target.

I tried to match source code style as much as possible - feel free to suggest any changes.


HowardHinnant:
Merged: https://github.com/HowardHinnant/date/commit/16439a8ce22d2d94d1666c1f2726adc34455cf27

Thanks!

PR #235, opened 2017-11-16
still some patches to build on Windows with clang in MSVC

lovemycatt:
date:: scope added to several floor<>() and round<>() calls to build on Windows with MSVC v140_clang_c2 and LLVM-vs2014.

This kind of problem only arises when instantiating template functions, therefore I am not sure if there are further calls that may need date:: scope...

### DIFF
--- a/include/date/date.h
+++ b/include/date/date.h
@@ -1098,7 +1098,7 @@ trunc(const std::chrono::duration<Rep, Period>& d)
 }
 
 #ifndef HAS_CHRONO_ROUNDING
-#  if defined(_MSC_FULL_VER) && _MSC_FULL_VER >= 190023918
+#  if defined(_MSC_FULL_VER) && (_MSC_FULL_VER >= 190023918 || (_MSC_FULL_VER >= 190000000 && defined (__clang__)))
 #    define HAS_CHRONO_ROUNDING 1
 #  elif defined(__cpp_lib_chrono) && __cplusplus > 201402 && __cpp_lib_chrono >= 201510
 #    define HAS_CHRONO_ROUNDING 1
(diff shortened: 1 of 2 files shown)


HowardHinnant:
Can this instead be fixed by adjusting the definition of `HAS_CHRONO_ROUNDING`?  This should be defined as `1` if the std::lib provides it, and otherwise `0`.

lovemycatt:
Thanks for the hint.
Actually when using clang under MSVC, for some reason the `_MSC_FULL_VER` value is reset to 1900**00000**, therefore the `HAS_CHRONO_ROUNDING` macro gets defined as `0`.
I'll fix the definition according to `__clang__` macro as well, clean up some code and re-push the branch for the PR.

HowardHinnant:
Ok, thanks.  Funny, when I got up this morning, I was thinking to myself that it probably could not be fixed that way.  :-)  But I haven't dug into it. Thanks for doing that!

lovemycatt:
I just re-pushed the branch.
This should work (at least it does here with VS2015 with v140_clang_c2, VS2015 with LLVM-vs2014 and Linux/GCC 6.1.1).

The only problem now is that `HAS_CHRONO_ROUNDING` will be misdefined to 1 if using clang or LLVM on MSVC with VS2015 with minor version < 23918: I believe this is a MSVC bugthat resets the minor version macro when using a different toolset than its default.
In such case, one should either install latest Service Pack or force `HAS_CHRONO_ROUNDING` to 0.

HowardHinnant:
Merged: https://github.com/HowardHinnant/date/commit/f4b12ab0233d5f31a915d52ac819eb88fcb5eb05

Thanks!

PR #244, opened 2017-11-24
Implementation of clock_cast function.

tomaszkam:
Implemented clock_cast<Target>(Source) function. For details please refer to commit messages.


include/date/tz.h
```diff
@@ -2183,42 +2197,224 @@ from_stream(std::basic_istream<CharT, Traits>& is, const CharT* fmt,
...
-template <class Duration>
-inline
-sys_time<typename std::common_type<Duration, std::chrono::seconds>::type>
-to_sys_time(const tai_time<Duration>& t)
+template<typename SourceClock, typename DestClock>
+struct clock_time_conversion
+{};
+
+template<>
+struct clock_time_conversion<std::chrono::system_clock, std::chrono::system_clock>
```
HowardHinnant: I think we can get rid of the identity conversions for `system_clock` and `utc_clock` as they can be handled by the generic identity conversion.  This simplifies us down to 3 `clock_time_conversion` specializations.

include/date/tz.h
```diff
@@ -2183,42 +2197,224 @@ from_stream(std::basic_istream<CharT, Traits>& is, const CharT* fmt,
...
+
+  // Check if Clock has to_sys method accepting TimePoint with given duration const& and returning sys_time
+  // If so has nested type member equal to return type to_sys.
+  template<typename Clock, typename Duration, typename = void>
+  struct return_to_sys 
+  {};
+
+  template<typename Clock, typename Duration>
+  struct return_to_sys<Clock, Duration, decltype(Clock::to_sys(std::declval<std::chrono::time_point<Clock, Duration> const&>()), void())>
+    : return_clock_time<std::chrono::system_clock, typename std::decay<decltype(Clock::to_sys(std::declval<std::chrono::time_point<Clock, Duration> const&>()))>::type>
```
HowardHinnant: Can we get rid of the `decay` trait here?  I can't imagine `to_sys` would not return by value, and if it did, it might be good to catch it as an error.

include/date/tz.h
tomaszkam: This would make ```clock_time_conversion<utc_clock, utc_clock>``` ambigous, as following candidates would be equal matches:
```
clock_time_conversion<Clock, Clock>
clock_time_conversion<Clock, utc_clock>
clock_time_conversion<utc_clock, Clock>
```
The same goes for ``std::chrono::system_clock``.

tomaszkam: At least on gcc-6.2 they are ambiguous (and I tend to agree with this interpretation).

HowardHinnant: Interesting.  Looks like we found a difference in behavior between gcc and clang.  We'll keep 'em and let the compilers work at that difference.

include/date/tz.h
tomaszkam: Just removing decay would lead to eliminating overload, which does not necessary means producing an error - example may select `utc_clock` conversion path, like in case of `amb1_clock` example.
The alternative would be to put an ``static_assert`` in `return_clock_time` that would check if TimePoint is not an reference.

tomaszkam: I like the idea. I will change `return_clock_time<C, TP>` to include an `static_assert` on `TP` being the same as `std::chrono::time_point<C, typename TP::duration>`. So havin `to_sys` returning something different than `sys_time` will be hard error, instead of SFINAE.

tomaszkam: Updated in next commit.

include/date/tz.h
```diff
@@ -2183,42 +2197,224 @@ from_stream(std::basic_istream<CharT, Traits>& is, const CharT* fmt,
...
+struct clock_time_conversion
+{};
+
+template<>
+struct clock_time_conversion<std::chrono::system_clock, std::chrono::system_clock>
 {
-    return to_sys_time(to_utc_time(t));
-}
+  template <class Duration>
+  auto operator()(const std::chrono::time_point<std::chrono::system_clock, Duration>& st)
```
HowardHinnant: `const`?

include/date/tz.h
```diff
@@ -2183,42 +2197,224 @@ from_stream(std::basic_istream<CharT, Traits>& is, const CharT* fmt,
...
-inline
-sys_time<typename std::common_type<Duration, std::chrono::seconds>::type>
-to_sys_time(const gps_time<Duration>& t)
+template<>
+struct clock_time_conversion<utc_clock, utc_clock>
 {
-    return to_sys_time(to_utc_time(t));
-}
+  template <class Duration>
+  auto operator()(const std::chrono::time_point<utc_clock, Duration>& st)
```
HowardHinnant: Make `const` here too?

include/date/tz.h
HowardHinnant: The parameter could be simplified to `const sys_time<Duration>&`

include/date/tz.h
HowardHinnant: The parameter can be simplified to `const utc_time<Duration>&`.

HowardHinnant:
We need to add this to the end of tz.h:
```c++
// Deprecated API

template <class Duration>
inline
sys_time<typename std::common_type<Duration, std::chrono::seconds>::type>
to_sys_time(const utc_time<Duration>& t)
{
    return clock_cast<std::chrono::system_clock>(t);
}

template <class Duration>
inline
sys_time<typename std::common_type<Duration, std::chrono::seconds>::type>
to_sys_time(const tai_time<Duration>& t)
{
    return clock_cast<std::chrono::system_clock>(t);
}

template <class Duration>
inline
sys_time<typename std::common_type<Duration, std::chrono::seconds>::type>
to_sys_time(const gps_time<Duration>& t)
{
    return clock_cast<std::chrono::system_clock>(t);
}


template <class Duration>
inline
utc_time<typename std::common_type<Duration, std::chrono::seconds>::type>
to_utc_time(const sys_time<Duration>& t)
{
    return clock_cast<utc_clock>(t);
}

template <class Duration>
inline
utc_time<typename std::common_type<Duration, std::chrono::seconds>::type>
to_utc_time(const tai_time<Duration>& t)
{
    return clock_cast<utc_clock>(t);
}

template <class Duration>
inline
utc_time<typename std::common_type<Duration, std::chrono::seconds>::type>
to_utc_time(const gps_time<Duration>& t)
{
    return clock_cast<std::chrono::system_clock>(t);
}


template <class Duration>
inline
tai_time<typename std::common_type<Duration, std::chrono::seconds>::type>
to_tai_time(const sys_time<Duration>& t)
{
    return clock_cast<tai_clock>(t);
}

template <class Duration>
inline
tai_time<typename std::common_type<Duration, std::chrono::seconds>::type>
to_tai_time(const utc_time<Duration>& t)
{
    return clock_cast<tai_clock>(t);
}

template <class Duration>
inline
tai_time<typename std::common_type<Duration, std::chrono::seconds>::type>
to_tai_time(const gps_time<Duration>& t)
{
    return clock_cast<tai_clock>(t);
}


template <class Duration>
inline
gps_time<typename std::common_type<Duration, std::chrono::seconds>::type>
to_gps_time(const sys_time<Duration>& t)
{
    return clock_cast<gps_clock>(t);
}

template <class Duration>
inline
gps_time<typename std::common_type<Duration, std::chrono::seconds>::type>
to_gps_time(const utc_time<Duration>& t)
{
    return clock_cast<gps_clock>(t);
}

template <class Duration>
inline
gps_time<typename std::common_type<Duration, std::chrono::seconds>::type>
to_gps_time(const tai_time<Duration>& t)
{
    return clock_cast<gps_clock>(t);
}
```
I'll remove these functions from the documentation, and add `clock_cast`, and `clock_time_conversion`.

tomaszkam:
Do we want to keep `clock_time_conversion` trait name? Any suggestions?

HowardHinnant:
I've been working on that.  `clock_convert_from_to` is the best I've come up with, and I haven't convinced myself that I like it better than `clock_time_conversion`.

HowardHinnant:
I'm working on the docs now.  I'll want to get that ok'd by both of us prior to pushing this work to master.  I sometimes find that the act of documentation reveals bugs or areas for improvement.  I'll run them by you when I have a first draft.

include/date/tz.h
HowardHinnant: Hmmm.. Shouldn't `clock_time_conversion<Clock, utc_clock>` be SFINAE'd out if `Clock::to_utc(utc_time<Duration>)` doesn't exist?

tomaszkam: The design behing the `clock_time_conversion<S, D>` is that the specialization is always valid, and only `operator()` is SFINAE'd. This apply to generic specialization (that does not have it at all), but also to `clock_time_conversion<C, C>` that will be valid for `C` being `int`, but the `operator()` will accept only `time_point<C>` - potentially this could be an identity.

Before, this specialization (`<S, utc_clock>`, `<utc_clock, D>`, `<S, sys_cloc>`, `<sys_clock, D>`) were SFINAE'd, but this workaround for compiler making code ill-formed if `S` was not having specific function (like `from_sys`). I have added the `return_to_sys` (and other) traits to avoid needs to SFINAE specialization and make them consistent.

Finally, there was not easy answer, what `Duration` should be used to SFINAE whole trait.

HowardHinnant:
What do you think about reversing the arguments to `clock_time_conversion` to `clock_time_conversion<DestClock, SourceClock>`, to make the order consistent with the order in `clock_cast`?

tomaszkam:
I was thinking a lot about the order of argument for the `clock_time_conversion`, but I found the order _from_, _to_ to be most natural one - the other example of similar usage would be `is_convertible` trait.

tomaszkam:
However, if you would prefer reverse one, I would be fine to change it.

HowardHinnant:
Merged: https://github.com/HowardHinnant/date/commit/543315b70096120436d5e9907946385a23718cce

Issue #18, opened 2015-10-21
CMake build


davidhunter22:
Would be nice to have a CMake based build. If you OK with this I will submit a pull request

HowardHinnant:
Is this not overkill for a single source?

davidhunter22:
Well a single source and a bunch of test cases. My main interest here is Visual Studio which is a bit of a pain to recreate each time. I agree that command line builds for GCC and Clang are pretty easy to do manually. The Cmake stuff also tends to make people use a consistent set of compiler/linker options and macros if any.  For instance I tend to add the -Weverything flag on clang builds.

knedlsepp:
Actually the installation of the "tz"-library is quite a bit more complicated than just downloading a single source file.

HowardHinnant:
Agreed.  The installation has grown more complex over time.  cmake is beginning to look more attractive.

dorpxam:
Cmake have some mechanisms that can help to automatize the download and install of the timezone database directly by using cmake before the build (not need to curl library installed) and avoid to use fixed path (in tz code) where to install the files (see my Windows Build recent message).

SylvainCorlay:
I was looking at making a conda package for this library and having CMake support (placing the `projectConfig.cmake` file in the right location) would really help.

ZaMaZaN4iK:
I am creating Conan package for the library. And would be great, if author will add CMake support. Without CMake i can't easily create package.

beached:
Is it just waiting for someone to do it?  I have one that should be close that I use for my repacked version of the library I use
https://github.com/beached/date/blob/master/CMakeLists.txt

beached:
Yay! :)

jkhoogland:
I had look at the CMakeLists.txt.

It has a hardcoded c++14 defined.
```
set_property(TARGET tz PROPERTY CXX_STANDARD 14)
```
I would suggest to add an option to select c++11, 14 or 17.

Also it is probable good to make the building of the tests optional.

r-darwish:
@jkhoogland I'm not sure I understand the reason for selecting `CXX_STANDARD`. I think that `CXX_STANDARD` should set to the minimum required by the library. If the library doesn't use C++17, then compiling it with `-std=c++17` shouldn't change anything.

HowardHinnant:
[...]

PR #245, opened 2017-11-25
Create CMakeList.txt

beached:
This is an implementation of cmake config relating to #18


HowardHinnant:
Thanks for your work on this.  Not ignoring, just preoccupied with another P/R at the moment.  Might be a few days before I respond.

Feel free to keep pushing changes to this P/R as opposed to closing and reopening.  Github will update this P/R with each of your pushes to `patch-2`.

beached:
Sorry about that, yeah I had an issue and normally would just do that.

beached:
I think I got it worked out
you can set the variables USE_SYSTEM_TZ_DB if you do not want to download from iana
```
cmake -DUSE_SYSTEM_TZ_DB=ON ..
```
If you want to download the iana db to the . folder you can set USE_TZ_DB_IN_DOT
```
cmake -DUSE_TZ_DB_IN_DOT ..
```
To run tests
```
cmake ..
make testit
```
If you want a shared library, static is default turn off BUILD_STATIC
```
cmake -DBUILD_STATIC=OFF ..
```
Otherwise the normal way of defining USE_AUTOLOAD, REMOTE_API... can be defined

HowardHinnant:
On macOS I get:
```
CMake Error at /Applications/CMake.app/Contents/share/cmake-3.9/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
  Could NOT find OpenSSL, try to set the path to OpenSSL root folder in the
  system variable OPENSSL_ROOT_DIR (missing: OPENSSL_INCLUDE_DIR)
```
I started to track it down and then realized that openssl isn't needed for this lib.

beached:
Isn’t it needed for curl to do https?

HowardHinnant:
I just link with `-lcurl`.  `curl` may need it, but at least on macOS that's a private implementation detail of `curl`.  I don't directly include or link against openssl.

beached:
Fixed and tested it on Ubuntu.  I checked for a feature flag of sorts but the CURL_VERSION_STRING in CMake only has the version number(e.g. 7.5.5).  If I find a system with curl that does not have ssl support(not sure it exists) I could document the failure so that it is obvious what is wrong.

HowardHinnant:
Merged: https://github.com/HowardHinnant/date/commit/f3741d68ff06f34a6bd71ea556fdc7257b7365c6

Thanks!

SylvainCorlay:
Quick feedback on the new CMakeList (sorry I was on the road earlier)

- Using the module `GNUInstallDirs` defines a number of variables for installation directories, which is a bit cleaner than using `INSTALLATION_PREFIX / include`.
- This will make cmake install the binaries in the right folder (lib, lib64) depending on the environment.
- Besides, a `ProjectConfig.cmake` would be really useful for library authors interested in depending on `date`.

PR #265, opened 2017-12-11
Add -fPIC when building static lib to enable linking with shared lib

tonywalker1:
When using libtz.a from a shared library, libtz must be compiled with -fPIC. My patch adds a new flag that optionally enables -fPIC when (and only when) building libtz as a static library. 

Notes: 
 * fPIC will be enabled by CMake automatically when building libtz as a shared library.
 * Therefore, another solution is to build libtz as a shared library when using it from a shared library.
 * For my use, I prefer building libtz as a static lib; however, I have no strong feelings on this patch should you choose to reject it. 

Thanks Howard and company for all of your hard work!


r-darwish:
I made a small project consisting of a shared library that tries to link with date statically. You are correct, the `-fPIC` flag is needed for this to work.

Basically, adding `set_property(TARGET tz PROPERTY POSITION_INDEPENDENT_CODE ON)` to my project's CMake solves it.

I tried to look whether other libraries that use CMake expose their own variable for enabling fPIC and found only one example. Most libraries do not do that.

We're pretty much just wrapping one variable with another, so I personally don't think the patch should be merged, and that people who wishes to build PIC should set the proper CMake flag in their project.

However, I will make sure that the Conan date package will have an option for controlling PIC.

tonywalker1:
Thanks. Like I said, I don't have strong feelings.

I am certainly not a CMake expert, but I think set_property() only works when building libtz as a subproject. If one builds it separately and installs it (e.g., in a container), then one needs to pass a flag in some way. However, enabling PIC on the CMake command-line is a fine solution (I think platforms that don't implement PIC ignore -fPIC). Following that line of reasoning, I think a number of Linux distributions build (or plan to build) as many binaries as possible with PIE and PIC for security. It might be worth considering making PIC the default for libtz. I think the only downside is a larger binary. That would remove one (admittedly small) problematic use case. Just a thought.

r-darwish:
I think that passing `-DCMAKE_POSITION_INDEPENDENT_CODE` to CMake will do if you're building it independently.

PR #266, opened 2017-12-11
Fix iOS build

kalliste:
The directory detection code for Mac introduced in 22a229af9106c032949f9a87a9a4fd9cc667a1ee makes us crash on iOS - this fixes the iOS version.

Note that we can't check #ifdef TARGET_OS_IPHONE here but have to do #if TARGET_OS_IPHONE because on Mac this value is defined and set to 0.


HowardHinnant:
Thanks!

PR #273, opened 2017-12-27
Add Travis file and add some fixes for libc++

r-darwish:
Despite these fixes the library still doesn't compile in Clang, which is why the clang job is
allowed to fail.

### DIFF
--- a/.travis.sh
+++ b/.travis.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+
+set -e
+
+if [[ ${CC} == "clang" ]]; then
+    CMAKE_FLAGS='-DCMAKE_CXX_FLAGS=-stdlib=libc++'
+fi
+
+cmake -DCMAKE_EXPORT_COMPILE_COMMANDS=ON ${CMAKE_FLAGS}
+make
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,7 @@
+language: cpp
+
+compiler:
+  - clang
+  - gcc
+
+script: ./.travis.sh


HowardHinnant:
I'm running with clang on macOS, and also have tip-of-trunk clang and libc++ available to test with here.  I'm not experiencing the problems you are addressing here.  Could you show the errors you are getting?

I'm not keen on adding Travis.  My experience with it has been less than encouraging with so many false failures that I've learned to ignore them.  I may well change my mind on this in the future, but today I am not ready to add Travis files.

The configuration macros such as `HAS_STRING_VIEW` are set up so that they are overridable from the command line.  They have suitable defaults for the case that they are not overridden from the command line.  The current default is based on `__cplusplus >= 201703`.  On what platform is this not working, and in what way is it not working?

r-darwish:
https://travis-ci.org/r-darwish/date/jobs/322189973

These are the errors that I got. I also managed to build the library with clang on my computer, but when using libstd++.

HowardHinnant:
I see, thanks for the link.  The travis build shows a combination of gcc-4.8 and `-std=gnu++1z`.  gcc-4.8 doesn't support enough std-lib beyond C++11 to be usable with date.h.  Specifically it lacks `std::enable_if_t` (and probably a bunch of other stuff).  But if one knocks things down to `-std=c++11`, and possibly adds `-DONLY_C_LOCALE` (not positive that is required), then I _believe_ one can use gcc-4.8.

r-darwish:
I modified the pull request to force linking with libc++ when clang is used. Everything seems to work now.

HowardHinnant:
Thanks.  But for at least a little while longer I'm going to maintain this as a Travis-free zone.

PR #274, opened 2017-12-31
Allow specifying cxx standard to target

beached:
Not everyone can use C++17 even with compilers that support it


HowardHinnant:
Thanks!

PR #277, opened 2018-01-05
Minutes value mishandled when parsing with negative offset 

scobei:
Parsing "2017-01-01T12:30:00-03:30" results in an offset of -150 minutes when it should be -210.


HowardHinnant:
Merged: https://github.com/HowardHinnant/date/commit/2b6ee6378c62b2f10b1515ea39bcb49579b56034

Thanks!

PR #278, opened 2018-01-07
Allow disabling of testing

beached:
Setting ENABLE_TESTING=OFF will disable testing

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,9 +7,9 @@ find_package( Threads REQUIRED )
 option( USE_SYSTEM_TZ_DB "Use the operating system's timezone database" OFF )
 option( USE_TZ_DB_IN_DOT "Save the timezone database in the current folder" OFF )
 option( BUILD_TZ_STATIC  "Build a static version of library" ON )
-option( ENABLE_TESTING  "Enable unit tests" ON )
+option( ENABLE_DATE_TESTING  "Enable unit tests" ON )
 
-if( ENABLE_TESTING )
+if( ENABLE_DATE_TESTING )
 	enable_testing( )
 endif( )
 
@@ -24,7 +24,7 @@ endfunction( )
 print_option( USE_SYSTEM_TZ_DB )
 print_option( USE_TZ_DB_IN_DOT )
 print_option( BUILD_TZ_STATIC  )
-print_option( ENABLE_TESTING  )
+print_option( ENABLE_DATE_TESTING  )
 
 if( USE_SYSTEM_TZ_DB )
 	add_definitions( -DUSE_AUTOLOAD=0 )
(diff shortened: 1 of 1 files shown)


SlavSlavov:
When I tested my proposal I did the following (file attached as well) :
- used more specific option name like ENABLE_DATE_TESTING or something similar. Using ENABLE_TESTING is more general and might be used in other components as well.
- moved enable_testing just above add_custom_target testit and guarded everything from eneble_testing to the end.
[CMakeLists.txt](https://github.com/HowardHinnant/date/files/1610178/CMakeLists.txt)

HowardHinnant:
@SlavSlavov Do you approve this P/R?  Or are you requesting modifications?  Thanks.

SlavSlavov:
I would suggest to use the modifications I made. Thanks.

beached:
I have no problem with ENABLE_DATE_TESTING as the flag, updated

SlavSlavov:
Here is the output of "make testit" with your modifications:
```
[user@workstation date_build]$ make testit
[100%] Built target tz
*********************************
No test configuration file found!
*********************************
Usage

  ctest [options]

[100%] Built target testit
[user@workstation date_build]$ 
```
And with my modifications:
```
[user@workstation date_build]$ make testit
make: *** No rule to make target 'testit'.  Stop.
[user@workstation date_build]$ 
```

HowardHinnant:
So what is the right move for this PR?  Looking for more advice, thanks.

SlavSlavov:
I have provided a CMakeLists.txt file (attached in one of my comments) which I would like to be used. My understanding is that when testing is not enabled, the target "testit" should not be available at all. 
I am not sure if I can commit changes to this pull request so just use the file I attached.

HowardHinnant:
Thanks @SlavSlavov But I don't have confidence that your recommended CMakeLists.txt is taking the latest CMakeLists.txt into account.  For example is it purposefully or accidentally reverting changes made in https://github.com/HowardHinnant/date/commit/637e5d80075c09de62db1282356dad56b1565754 ?

SlavSlavov:
@HowardHinnant - sorry, this had to be a very simple change and I think it is going in the wrong direction. When I initially proposed it in #18, I proposed how to be done. When I added the file to this thread, the master branch was not yet updated. I'll post the steps to make the change here. All you need is to add an option which enables testing and at the end guards everything related to testing (not the contents of each function as the commits in this PR suggest). If you think this makes sense and can be done easily I would appreciate having it. If not just delete the PR.


Remove the line ```enable_testing( )``` which is at the beginning of the file

Add the following option at the end of your options list:
```
option( ENABLE_DATE_TESTING  "Enable unit tests" ON )
```

Add the following option at the end of your print options list
```
print_option( ENABLE_DATE_TESTING )
```

Add the following just before the line ```add_custom_target( testit COMMAND ${CMAKE_CTEST_COMMAND} )``` and indent everything to the end of the file

```
if ( ENABLE_DATE_TESTING )

    enable_testing( )
```

At the end of the file add ```endif( )``` 

Now everything related to testing is guarded and should look like this:


```
if ( ENABLE_DATE_TESTING )

    enable_testing( )

    add_custom_target( testit COMMAND ${CMAKE_CTEST_COMMAND} )

    ....

    endforeach( )
endif( )
```

HowardHinnant:
@SlavSlavov @beached For your review:  https://github.com/HowardHinnant/date/pull/291

HowardHinnant:
Addressed by https://github.com/HowardHinnant/date/commit/4ada98d247b5b68a061efdcf4b0a276b1e966e57

PR #279, opened 2018-01-09
Moved ios support into source directory.

lava:
By common convention, the include/ directory contains the public
interface of a library. The `ios.h` header is only included by
`tz.cpp` and defines an internal helper function, so it was moved
into src/. The `ios.mm` file actually contains C++ source code, so it
probably was not supposed to be in include/ in the first place.

In addition, the include in `tz.cpp` was changed to only include
the `ios.h` header when compiling on apple platforms, as the
documentation states that only the four files date.h, tz.h, tz_private.h,
and tz.cpp should be required to use the time zone library.


HowardHinnant:
I'm in agreement that ios.mm belongs in src.  However I would like to keep ios.h in include.  tz_private.h also remains in include (another internal-only header).

lava:
Ok, updated as requested.

HowardHinnant:
Merged: https://github.com/HowardHinnant/date/commit/2acf403bcda303b27acc092c5fe4e165ec4f6614

Thanks!

PR #280, opened 2018-01-10
Export a CMake config file

remyabel:
This will allow users to import the project via CMake methods, i.e `find_package` rather than doing it manually.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -89,9 +89,14 @@ target_include_directories(date_interface INTERFACE
     $<INSTALL_INTERFACE:include>
 )
 
+if(WIN32 AND NOT CYGWIN)
+    set(DEF_INSTALL_CMAKE_DIR CMake)
+else()
+    set(DEF_INSTALL_CMAKE_DIR lib/cmake/date)
+endif()
 
 install( TARGETS date_interface EXPORT DateConfig )
-install( EXPORT DateConfig DESTINATION lib/cmake/date )
+install( EXPORT DateConfig DESTINATION ${DEF_INSTALL_CMAKE_DIR} )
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -94,6 +94,8 @@ target_include_directories(date_interface INTERFACE
 )
 
 
+install( TARGETS date_prj EXPORT DateConfig )
```
SlavSlavov: Have you tried to generate project with these changes? When I try I get the following error:
```
CMake Error at CMakeLists.txt:96 (install):
  install TARGETS given target "date_prj" which does not exist in this
  directory.
```
There is no target date_prj in this file.

CMakeLists.txt
```diff
@@ -94,6 +94,8 @@ target_include_directories(date_interface INTERFACE
...
+install( EXPORT DateConfig DESTINATION lib/cmake/date )
```
SlavSlavov: If this will be used on systems different from Linux, the destination must be set accordingly. ```find_package``` searches in different directories on different systems:
https://cmake.org/cmake/help/latest/command/find_package.html#command:find_package
An example to set it would be:
```
if(WIN32 AND NOT CYGWIN)
    set(DEF_INSTALL_CMAKE_DIR CMake)
else()
    set(DEF_INSTALL_CMAKE_DIR lib/cmake/SeqsiBuild)
endif()
```

CMakeLists.txt
remyabel: Thanks.

HowardHinnant:
Merged https://github.com/HowardHinnant/date/commit/362cd8f27efbb5e8546f792b82e63c6fffa2ea0e

PR #281, opened 2018-01-10
Use target_compile_definitions

remyabel:
It's generally preferred to use [`target_compile_definitions`](https://floooh.github.io/2016/01/12/cmake-dependency-juggling.html) over `add_definitions`.


HowardHinnant:
Merged: https://github.com/HowardHinnant/date/commit/637e5d80075c09de62db1282356dad56b1565754

PR #286, opened 2018-01-12
implement current_zone() for iOS using native API from iOSUtils

allidev:
Use native system call to retrieve current zone info.


HowardHinnant:
Thank you so much for doing this again!

PR #295, opened 2018-02-14
[tz.h] Simpler workaround for MSVC

alexkaratarakis:
This is an alternative to having the `#ifdef` with specific `_MSC_VER` and bumping it every version.

### DIFF
--- a/include/date/tz.h
+++ b/include/date/tz.h
@@ -2197,8 +2197,6 @@ from_stream(std::basic_istream<CharT, Traits>& is, const CharT* fmt,
     return is;
 }
 
-#if !defined(_MSC_VER) || _MSC_VER > 1913
-
 // clock_time_conversion
 
 template <class DstClock, class SrcClock>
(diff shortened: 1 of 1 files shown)


alexkaratarakis:
Oh, right, I should mention I am in the VS team and we are aware of the issue. :)

We are suggesting this workaround because the fix is unlikely to be included with the default compiler flags (it will likely require /permissive-) so the workaround will still be required.

Furthermore, the old workaround is "hiding" ~250 lines of code, whereas the new one is isolated to the problematic spots. Therefore, with the new workaround, those lines will be getting tested (we are indeed testing `date` in our internal CIs).

HowardHinnant:
Even more interesting, thanks!

HowardHinnant:
Merged: https://github.com/HowardHinnant/date/commit/a1ceec19fed2196ff4e22929100b221160651aa6

Thanks!

PR #297, opened 2018-02-21
Fix VS compilation when there is a `string` type in the global scope

mwu-tow:
I'm introducing the date library to a legacy codebase. It has defined `string` class (other than `std::string`) in a global scope, making the date header to fail to compile.

The following:
```
struct string {}; 
#include <date/date.h>
``` 

fails to compile VS 2017.5 with `error C2872: 'string': ambiguous symbol`.

I fixed that by explicitly qualifing the `std::string` usage with `std` namespace. 
I see there are other symbols which might suffer from the same issue (like `numpunct`) but for now I fixed the one that stopped me from using the library.


HowardHinnant:
Thanks!

There are more issues like this when compiling with other flags (e.g. `DONLY_C_LOCALE`).  I've decided to address them only on an "as needed" basis.

PR #301, opened 2018-03-02
Fix case of generated dateConfig so it will work on a case sensitive …

dsmyth:
Fix case of generated dateConfig so it will work on a case sensitive filesystem.

This way it will work with cmake find_package on Linux.


HowardHinnant:
Thanks!

PR #304, opened 2018-03-06
isspace, isalnum are undefined for signed chars?

ChrisBFX:
Full disclosure: I'm not that confident in my change, as I never really worked with those functions (especially I've no idea what happens if use wide chars as inputs?).

http://en.cppreference.com/w/cpp/string/byte/isspace says:
> the behavior of std::isspace is undefined if the argument's value is neither representable as unsigned char nor equal to EOF. To use these functions safely with plain chars (or signed chars), the argument should first be converted to unsigned char

I noticed this when using from_stream with a format string which contained utf8 characters under VisualStudio2017 which raised a debug assertion.

### DIFF
--- a/include/date/date.h
+++ b/include/date/date.h
@@ -6855,7 +6855,7 @@ from_stream(std::basic_istream<CharT, Traits>& is, const CharT* fmt,
                 }
                 else  // !command
                 {
-                    if (isspace(*fmt))
+                    if (isspace(static_cast<unsigned char>(*fmt)))
                         ws(is); // space matches 0 or more white space characters
                     else
                         read(is, *fmt);


include/date/date.h
```diff
@@ -6838,7 +6838,7 @@ from_stream(std::basic_istream<CharT, Traits>& is, const CharT* fmt,
                                 auto wc = Traits::to_char_type(i);
                                 auto c = static_cast<char>(wc);
                                 // is c a valid time zone name or abbreviation character?
-                                if (!(CharT{1} < wc && wc < CharT{127}) || !(isalnum(c) ||
+                                if (!(CharT{1} < wc && wc < CharT{127}) || !(isalnum(i) ||
```
HowardHinnant: I think this one is ok because of the preceeding `!(CharT{1} < wc && wc < CharT{127})`

include/date/date.h
```diff
@@ -6879,7 +6879,7 @@ from_stream(std::basic_istream<CharT, Traits>& is, const CharT* fmt,
                 }
                 else  // !command
                 {
-                    if (isspace(*fmt))
+                    if (isspace(Traits::to_int_type(*fmt)))
```
HowardHinnant: I don't think this one is fixed.  Perhaps `static_cast<unsigned char>(*fmt)`.  That isn't perfect, but it would at least avoid UB.

ChrisBFX: Changed it to static_cast, which is actually what char_traits<char>::to_int_type does (at least in libstdc++ and msvc).

PR #308, opened 2018-03-13
Use proper lib directory when installing on UNIX

evsh:
Linux distributions use lib, lib32, and lib64 directories for library files,
symlinking them where needed. Let's follow distirbution rules by
utilising CMake module GNUInstallDirs.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,8 @@ cmake_minimum_required( VERSION 3.0.2 )
 
 project( date_prj )
 
+include( GNUInstallDirs )
+
 find_package( Threads REQUIRED )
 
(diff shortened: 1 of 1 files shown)


HowardHinnant:
Will this work on Windows?

evsh:
`if( UNIX AND NOT APPLE )` condition should fallback to the `else` branch on Windows and MacOS. The fallback branch uses 'lib' as it is now. I did not test this change neither on Windows not on MacOS, but I use the same code in other projects and it works fine with those OSes.

evsh:
Realised that things can be simplified avoiding conditions.

HowardHinnant:
Merged: https://github.com/HowardHinnant/date/commit/f105595f04b0be26435df64f3e5b22fbf0c8c46c

Thanks!

PR #312, opened 2018-03-15
Validate TZdata failed with OS TZdata

hjmallon:
`USE_OS_TZDB` was not defined in test programs so the validate tz data test failed on macOS with `USE_SYSTEM_TZ_DB=On`.


hjmallon:
This also causes he clock tests to fail to build. Before they would segfault instead.

Issue #313, opened 2018-03-15
configuration header


cbeck88:
Currently, when we install the tz library using cmake, we select a bunch of configuration options, for example:

```
cmake -DENABLE_DATE_TESTING=NO \
             -DUSE_SYSTEM_TZ_DB=YES \
             -DTZ_CXX_STANDARD=14 \
             -DBUILD_TZ_STATIC=YES \
             -DCMAKE_CXX_FLAGS=-fPIC \
             -DCMAKE_INSTALL_PREFIX=/usr \
             .. \
```

However, we can't then directly use the headers as is, we also have to add a bunch of defines to the build system, per installation instructions here: https://howardhinnant.github.io/date/tz.html#Installation

So we end up writing our own configuration header like so:

```
#pragma once

#define USE_OS_TZDB 1

#include "date/date.h"
#include "date/tz.h"
```

and we tell developers that it is a bug if they include `date/tz.h` directly in their code and not using our in-tree configuration header.

This also means that the configuration header in-tree needs to be kept in sync with the code that is invoking cmake.

It would be nicer if, for example, the `date` library cmake would write its own configuration header, e.g. install `tz_config.h` alongside `tz.h`. And `tz.h` would itself include this header, which would contain the defines that we selected at installation time. This is the pattern followed by other common open source libraries like `libpng` and `freetype`. And it avoids the possibility of ODR violations and crashes if the installation options and the options in my hand-written header above get out of sync.

Would you guys be interested in a Pull Request that introduces such a configuration header into your cmake?

jeremy-coulon:
I wish there was other CMake options in order to be able to load TZ database from a custom location.
Currently options are:
* USE_SYSTEM_TZ_DB
* USE_TZ_DB_IN_DOT

I would like another option (e.g. TZ_AUTO_DOWNLOAD).

Then I could set all options to OFF and specify my tzdata installation at runtime with "void set_install(const std::string& s);" API.

jeremy-coulon:
```
if( USE_SYSTEM_TZ_DB AND NOT WIN32 )
        # cannot set USE_OS_TZDB to 1 on Windows
        target_compile_definitions( tz PUBLIC -DUSE_OS_TZDB=1 )
else( )
        target_compile_definitions( tz PUBLIC -DUSE_OS_TZDB=0 )
endif( )
if( TZ_AUTO_DOWNLOAD )
        target_compile_definitions( tz PRIVATE -DAUTO_DOWNLOAD=1 )
[...]

PR #315, opened 2018-03-19
Add a configuration header (issue #313)

cbeck88:
This makes it easier to install and use the tz library correctly,
since the correct defines, which were used when tz was compiled
will be set by the configuration header which is automatically
included by tz.h

This follows the pattern set by common open source libraries like
libpng and freetype.

This commit also fixes a bug where the define USE_AUTOLOAD changed
name to AUTO_DOWNLOAD but the cmake was not updated correspondingly.

This was caught because we added

at the top of the tz.h header after we include the configuration
header.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,35 +40,44 @@ set( HEADER_FILES
 	${HEADER_FOLDER}/date/tz_private.h
 )
 
-if( BUILD_TZ_STATIC )
-	add_library( tz STATIC ${HEADER_FILES} ${SOURCE_FOLDER}/tz.cpp )
-else( )
-	add_library( tz SHARED ${HEADER_FILES} ${SOURCE_FOLDER}/tz.cpp )
-endif( )
-
 if( USE_SYSTEM_TZ_DB )
-	target_compile_definitions( tz PRIVATE -DUSE_AUTOLOAD=0 )
-	target_compile_definitions( tz PRIVATE -DHAS_REMOTE_API=0 )
+	set(AUTO_DOWNLOAD 0)
+	set(HAS_REMOTE_API 0)
+
 	# cannot set USE_OS_TZDB to 1 on Windows
 	if( NOT WIN32 )
-		target_compile_definitions( tz PUBLIC -DUSE_OS_TZDB=1 )
+		set(USE_OS_TZDB 1)
+	else ( )
+		set(USE_OS_TZDB 0)
 	endif( )
 else( )
-	target_compile_definitions( tz PRIVATE -DUSE_AUTOLOAD=1 )
-	target_compile_definitions( tz PRIVATE -DHAS_REMOTE_API=1 )
-	target_compile_definitions( tz PUBLIC -DUSE_OS_TZDB=0 )
+	set(AUTO_DOWNLOAD 1)
+	set(HAS_REMOTE_API 1)
+	set(USE_OS_TZDB 0)
+
 	find_package( CURL REQUIRED )
 	include_directories( SYSTEM ${CURL_INCLUDE_DIRS} )
 	set( OPTIONAL_LIBRARIES ${CURL_LIBRARIES} )
+endif( )
+
+if( NOT TZ_CXX_STANDARD )
+	set( TZ_CXX_STANDARD 17 )
+endif( )
+
+set(TZ_CONFIG_H_PATH ${CMAKE_CURRENT_BINARY_DIR}/${HEADER_FOLDER}/date/tz_config.h)
+configure_file(tz_config.h.in ${TZ_CONFIG_H_PATH} @ONLY)
+
+
+if( BUILD_TZ_STATIC )
+	add_library( tz STATIC ${HEADER_FILES} ${TZ_CONFIG_H_PATH} ${SOURCE_FOLDER}/tz.cpp )
+else( )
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -40,35 +40,44 @@ set( HEADER_FILES
...
 
+if( NOT TZ_CXX_STANDARD )
+	set( TZ_CXX_STANDARD 17 )
+endif( )
+
+set(TZ_CONFIG_H_PATH ${CMAKE_CURRENT_BINARY_DIR}/${HEADER_FOLDER}/date/tz_config.h)
+configure_file(tz_config.h.in ${TZ_CONFIG_H_PATH} @ONLY)
+
+
+if( BUILD_TZ_STATIC )
```
hjmallon: We can simplify this. See https://cmake.org/cmake/help/v3.0/command/add_library.html.

If `BUILD_TZ_STATIC` just sets `BUILD_SHARED_LIBS` to the correct value and type is not explicitly set in this call.

hjmallon: Could definitely be a separate change though.

hjmallon: I put that change in here for reference: https://github.com/HowardHinnant/date/pull/316

cbeck88: One thing to consider is, if tz.cpp is e.g. compiled at C++11 standard, and then tz.hpp is used in a compilation unit that compiles at C++17 standard, is that going to cause ODR violations? Especially if e.g. `HAS_STRINGVIEW` is true in one version and false in the other?

It might be a good idea to make standard detection part of the configuration header, for tz lib, to avoid such problems. It's likely going to be more work than this PR was though, up to you guys if it's worth it.

hjmallon: Currently a utility using tz.h with c++ without `std::string_view` will use/link functions with `std::string` arguments. That will fail if the library was built with `std::string_view` arguments unless there is some bridging malarkey in the string_view stuff that I do not know about (possible).

Perhaps the library should make symbols for `std::string` and `std::string_view` if possible? Then it could be linked against utilities using either (the switch would be in the header).

Using the default `CMAKE_CXX_STANDARD` does mean that anyone who ties this library into their build system (with `add_subdirectory`) will automatically build it with the same `CXX_STANDARD` as the outer utility.

cbeck88: Yeah, I don't disagree that `CMAKE_CXX_STANDARD` should be used.

I'm looking through the code now...

I guess I'm worried about something like this:

https://github.com/HowardHinnant/date/blob/master/include/date/tz.h#L101

```
#ifndef HAS_DEDUCTION_GUIDES
#  if __cplusplus >= 201703
#    define HAS_DEDUCTION_GUIDES 1
#  else
#    define HAS_DEDUCTION_GUIDES 0
#  endif
#endif // HAS_DEDUCTION_GUIDES
```

I'd rather have `cmake` figure out this value, using `CMAKE_CXX_STANDARD` at the time that date is configured, and then write a variable `HAS_DEDUCTION_GUIDES` into `tz_config.h`, so that it has consistent configuration both when it is built and when it is consumed.

include/date/tz.h
```diff
@@ -43,20 +43,18 @@
 // required. On Windows, the names are never "Standard" so mapping is always required.
 // Technically any OS may use the mapping process but currently only Windows does use it.
 
+#include "tz_config.h"
+
 #ifndef USE_OS_TZDB
-#  define USE_OS_TZDB 0
+#  error "Missing define expected from configuration header"
 #endif
```
HowardHinnant: Sorry, can't do these changes to tz.h.  CMake is just _a_ build system.  It is not _the_ build system.

I want to keep it easy to configure this lib from the command line by using `-DMACRO=VALUE`, and to have reasonable default values for the macros.

cbeck88: I see, but are you still okay with `#include "tz_config.h"` then?

cbeck88: The thing I'm concerned about is, when I use this in a larger project, if we build `tz.cpp` with one set of macros, and include it in a larger project using a different set of macros, we just get undefined behavior, and this caused a bunch of crashes in our firmware.

When cmake chooses values for those macros and then compiles tz.cpp and installs it, I want to still have a source of truth for what those values are, otherwise, it's a bit risky to link into the project

Are you okay with an `if has_include "tz_config.hpp"` or whatever the right syntax for that is? I think it might be possible on gcc and clang, I don't know if it's standard, I can't remember

HowardHinnant: `__has_include` was standardized in C++17.  This project still has to support C++11 and C++14.  If there's a way to do that for clang, gcc and VC++, I'm fine with that.  I haven't investigated whether there is such a way or not.

I'm ok with adding a config file.  But I want to keep the ability to build from the command line for projects:

*  That just use date.h.
*  That use tz.h.
*  That just use iso_week.h, islamic.h, julian.h, or ptz.h.

CMake support has to be a pure add-on to this lib.  Also, the less files the better.  In many places I've already advertised that if you just need civil calendar support, date.h is all you need.  It would be nice to keep that.

cbeck88: Can "build from command line" procedure include "copy `config.h.in` from the root of the library and adjust as needed for your project, and put it in the include path"?

Or you want a different tool that does that?

I guess we could do something like pass one define `-DTZ_USE_CONFIG_HEADER` that causes it to include `tz_config.h`, and otherwise it doesn't do it. That would still be safer than trying to get all the configuration options right in two different places. Does that sound good to you?

PR #330, opened 2018-04-13
Add string_view support for Visual Studio 2017

qis:
Here is a small patch that adds `string_view` support for Visual Studio 2017. It should make future [vcpkg](https://github.com/Microsoft/vcpkg) integration easier.

Some tests in the `testit` CMake target failed, but they seem to have nothing to do with my changes:

```
13 - date_test_pass_range_test (Failed)
32 - date_test_pass_parse_test (Failed)
44 - date_test_pass_year_month_test (Failed)
45 - date_test_pass_year_month_day_test (Failed)
46 - date_test_pass_year_month_day_last_test (Failed)
47 - date_test_pass_year_month_weekday_test (Failed)
48 - date_test_pass_year_month_weekday_last_test (Failed)
61 - tz_test_pass_validate_test (Failed)
```

### DIFF
--- a/include/date/date.h
+++ b/include/date/date.h
@@ -31,7 +31,7 @@
 // We did not mean to shout.
 
 #ifndef HAS_STRING_VIEW
-#  if __cplusplus >= 201703
+#  if __cplusplus >= 201703 || (defined(_MSVC_LANG) && _MSVC_LANG >= 201703L)
 #    define HAS_STRING_VIEW 1
 #  else
 #    define HAS_STRING_VIEW 0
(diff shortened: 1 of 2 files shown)


HowardHinnant:
I _think_ this has been resolved, but I am not 100% positive because I don't have VS to test with.  Please reopen if there is still a problem at tip-of-master.  Thanks.

qis:
@HowardHinnant By now this is definitely the wrong solution, but your check if `string_view` is available still fails in MSVC. This would be correct:
```
#ifndef HAS_STRING_VIEW
#  if __cplusplus >= 201703 || (defined(_MSVC_LANG) && _MSVC_LANG >= 201703L)
#    define HAS_STRING_VIEW 1
#  else
#    define HAS_STRING_VIEW 0
#  endif
#endif  
```

qis:
@HowardHinnant Thank you **very** much for this change from the bottom of my heart!

PR #336, opened 2018-05-06
Fix ambiguity in arithemtic with duration convertible to years.

tomaszkam:
Implemented fix for the ambiguity of arithmetic with duration convertible to years that was reported by Tim Song. See commit descriptions for details.
Added test for the case.


tomaszkam:
1. Will do.
2. It would be fine if we will just remove the years overload and 
implement it directly.
     However I went that way we have currently, so we have a clear 
separation of:
      1) Functions performing actual arithmetic that are non-template.
      2) Template wrappers to resolve ambiguities.
     I can reduce the overload by one, but done this on purpose. Please 
let me know, if you just want to have two overloads.

3. The test work fine on g++-7 with stdlibc++. From what I checked on 
libc++, the duration constructors are no noexcept at all. So:
static_assert(noexcept(ym + months{1}));
was never true. Could you check the file without my changes (only 
decades should be ambiguous).

4. Will do. I noticed that you are putting just "".

5. I haven't separated them, because the declarations is consuming 80% 
of the code, and would need to be repated anyway.
     Please let me know, if you want to have it changed anyway.

6. Will do.

W dniu 06.05.2018 o 21:20, Howard Hinnant pisze:

HowardHinnant:
2.  I think I would rather minimize the number of signatures.  It adds up fast with all the overloads between +=, -=, and the binary operators already.

3.  I wouldn't mind if the chrono::durations gained conditionally noexcept constructors, but that's a fight for another day.  

> Could you check the file without my changes (only decades should be ambiguous).

Not sure what you mean, but I did test the overload set I suggested for +=, -= and it worked.

4.  Sounds good, thanks.

5.  Yes, separation would be good.  With us dropping back down to two sets instead of three, I believe you can just modify the already separated signatures.

Thanks much for doing this.

tomaszkam:
I have updated the test cases. Could you please recheck them?
Note that all operations invoking conversion of duration, will not be noexcept, so I commented them.

tomaszkam:
I mean for the noexcept - could you check that all asserts (except decads) works for you both before and after my changes. I have problems with running libc++ (missing xlocale error).

tomaszkam:
Per commit message:
As the conversion constructor from years to months is not marked
as noexcept, the expression is_noexcept_constructible<months, Duration const&>::value
was returning false, even if Duration was years.
Changed the condition, to check if we can convert either to
years of months.

The other option would be:
1) Reintroduce separate months operator for year_month_day*.
2) Drop noexcept.

HowardHinnant:
All tests are passing for me, and things look good for C++11 as well.  I like your current conditional noexcept specification.  I note that `std::is_nothrow_constructible<years, years const&>::value` is true, which means the current API doesn't change.

tomaszkam:
I have separated implementations from declarations for all classes. Also added parenthesis to NOEXCEPT_CONVERTIBLE_TO..., as g++ were complaining.

tomaszkam:
I have also removed the operator+/-=(moths) for year_month. The operator was just forwarding to operator+/-, so it could be replaced with single template.

tomaszkam:
I am not sure if you are waiting to see how much I could improve the pull, but the apporach is working:
I have refactored the test for the addition, to compile both in C++11 and C++14 modes, by defining macros that expand either to static_assert or plain assert.
Also fixed some issue in clock_cast test itself that made them incompatible with C++11: common_type_t and constexpr conversion from year_month_day to sys_days.

HowardHinnant:
Naw, just took the night off.  I'll probably look at this tonight. :-)

tomaszkam:
Have you had time to look at the changes?

HowardHinnant:
Merged: https://github.com/HowardHinnant/date/commit/328cecaa567e8b9da5dfcbcb4de9a550496bb55b

You'll need to `git reset --hard` to my master branch as I changed some whitespace and folded all of your commits into one.

Sorry for the slow response on this one.  Busy week.  Thanks for the work!

PR #341, opened 2018-05-25
Eliminate use of uninitialized offset.

memorizer:
Parsing of dates in format "2018-05-25T09:22:38.684811653Z" is broken. This date is parsed to 1805429794651672965 nanoseconds which is equivalent to GMT: Friday, March 19, 2027 4:16:34 AM

Reference code:
std::string strDate = "2018-05-25T09:22:38.684811653Z";
std::istringstream in{ strDate };

date::sys_time<std::chrono::nanoseconds> tp;
in >> date::parse("%FT%TZ", tp);


HowardHinnant:
Thanks!

PR #347, opened 2018-06-10
Local time support for utc/tai/gps/sys clock

tomaszkam:
This commit ads ability to create an time point at given date/time (2018-12-03 12:00) for all clocks. This is done by converting an local time instance to the clock time, using clock cast.

The only drawback is that the local time is not able to express an time during leap second insertion for UTC (there is no way to express 60s). So this need to be expressed manually.


HowardHinnant:
This is really nice work.  And thanks.  I'm happy with it, but I'm going to hold off merging for a few more hours, so take your time.

tomaszkam:
I am still on the fence with utc_clock. As we have an separate specialization for clock_time_conversion between utc_clock and local_t, I was thinking that maybe it would be better to expose something like:
std::pair<local_time, bool> to_local_with_leap();
utc_time from_local_with_leap();
Now, if we are inside with leap second insertion, we return true, and the local_time is equivalent to:
leap_insertion_time + time_during_leap (0-1s)

Now, the clock_time_conversion will provide current output (lowest unit before), but we could use it directly in I/O.

HowardHinnant:
Thanks!

Merged: https://github.com/HowardHinnant/date/commit/01978895055e4216a5c2303b67da102256812f06

PR #348, opened 2018-06-11
Add option to disable string_view

cdalvaro:
Also add -fPIC flag when compiling the shared version of the library.


cdalvaro:
Thank you very much!

PR #350, opened 2018-06-12
Remove usage of octal literals in tests

tomaszkam:
It was pointed out by my work colleague that use of numbers padded with zeros is a bad style - they are interpreted as octal literals, and works correctly (gives correct value), unless someone will want to configure September or 9th.

This only affects local_t test, that I have recently added.

### DIFF
--- a/test/clock_cast_test/local_t.pass.cpp
+++ b/test/clock_cast_test/local_t.pass.cpp
@@ -39,7 +39,7 @@ main()
 
    // utc leap second
    {
-     auto lu = local_days{2015_y/07/1_d} - milliseconds(1);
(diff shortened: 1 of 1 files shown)


HowardHinnant:
I wouldn't say bad style, just something you have to know.  You either get the right answer or a compile-time error (never an incorrect value at run time).  No big deal.  But as long as we're going to be consistent on this issue, search for "07" in local_t.pass.cpp (missed one).

tomaszkam:
I am well aware that number prefixed with 0 are octal literals, but that didn't stop me from following the habit to pad dates/months in dates.

//Please squash the commit on merge.

PR #351, opened 2018-06-16
Use string_literal for C++11 when possible

tomaszkam:
Merged get_units code for C++11 and C++14, now string_literal are used for all non-custom ratio suffix, that are at most 2 characters [except: decaseconds, minutes, microseconds in char].

This requires to special cases for string_literal, for C++11 mode:
1) Constructor for from string or length 2 (3 elements array)
2) Adding two string literals with size 1.


tomaszkam:
Accidentally removed the branch, sorry.

PR #352, opened 2018-06-16
Use string_literal for C++11 when possible

tomaszkam:
Merged get_units code for C++11 and C++14, now string_literal are used for all non-custom ratio suffix, that are at most 2 characters [except: decaseconds, minutes, microseconds in char].

This requires to special cases for string_literal, for C++11 mode:
1. Constructor for from string or length 2 (3 elements array)
2. Adding two string literals with size 1.

### DIFF
--- a/include/date/date.h
+++ b/include/date/date.h
@@ -7858,20 +7858,20 @@ operator+(const string_literal<CharT, 2>& x, const string_literal<CharT, 2>& y) 
   return string_literal<CharT, 3>({x[0], y[0], 0});
 }
 
-#if __cplusplus >= 201402  && (!defined(__EDG_VERSION__) || __EDG_VERSION__ > 411) \
-                           && (!defined(__SUNPRO_CC) || __SUNPRO_CC > 0x5150)
-
 template <class CharT1, class CharT2, std::size_t N1, std::size_t N2>
 CONSTCD14
 inline
-string_literal<std::conditional_t<sizeof(CharT2) <= sizeof(CharT1), CharT1, CharT2>,
+string_literal<typename std::conditional<sizeof(CharT2) <= sizeof(CharT1), CharT1, CharT2>::type,
                N1 + N2 - 1>
 operator+(const string_literal<CharT1, N1>& x, const string_literal<CharT2, N2>& y) NOEXCEPT
 {
-    using CharT = std::conditional_t<sizeof(CharT2) <= sizeof(CharT1), CharT1, CharT2>;
+    using CharT = typename std::conditional<sizeof(CharT2) <= sizeof(CharT1), CharT1, CharT2>::type;
     return string_literal<CharT, N1 + N2 - 1>{string_literal<CharT, N1>{x},
                                               string_literal<CharT, N2>{y}};
 }
+
+#if __cplusplus >= 201402  && (!defined(__EDG_VERSION__) || __EDG_VERSION__ > 411) \
+                           && (!defined(__SUNPRO_CC) || __SUNPRO_CC > 0x5150)
 
 template <class CharT,
           class = std::enable_if_t<std::is_same<CharT, char>{} ||
@@ -7956,34 +7956,6 @@ msl(std::ratio<N, D>) NOEXCEPT
     return msl(CharT{'['}) + msl<R::num>() + msl(CharT{']'});
 }
 
-template <class CharT>
-CONSTCD14
-std::enable_if_t
-<
-    std::is_same<CharT, char>{},
-    string_literal<char, 3>
->
-msl(std::micro) NOEXCEPT
-{
-    return string_literal<char, 3>{"\xC2\xB5"};
-}
-
-template <class CharT>
-CONSTCD14
-inline
-auto
-msl(std::deca) NOEXCEPT
-{
-    return string_literal<CharT, 3>{"da"};
-}
-
-template <class CharT>
-CONSTCD11
-auto
-get_units(std::ratio<60>)
-{
-    return string_literal<CharT, 4>{"min"};
-}
 
 #else  // __cplusplus < 201402 || (defined(__EDG_VERSION__) && __EDG_VERSION__ <= 411)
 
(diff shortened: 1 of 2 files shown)


tomaszkam:
I have also merged code for decaseconds, microseconds and minutes, as I realized that runtime constructing and concatenation string_literals, is no worse than doing the same thing for runtime sized string (that needs to handle SSO).

I preserved special case, that makes most of the cases constrexpr in C++11.

tomaszkam:
Changed get_untis that combine preffix with 's' to be C++11 constexpr, as 2 letter case is supported in C++11.

tomaszkam:
Please squash commits on merge.

tomaszkam:
For the last commits, I am not sure how do you feel about test testing implementation detail, but I found it to be only way to confirm that we get constexpr things.

tomaszkam:
Also, by adding two additional functions, we can make all standard suffix constexpr in C++11. Want to have your opinion.

HowardHinnant:
I'm getting this unit test failure.  I haven't tracked down the cause yet.
```
/Users/howardhinnant/Development/date
passed 1 tests in /Users/howardhinnant/Development/date/test
passed 8 tests in /Users/howardhinnant/Development/date/test/clock_cast_test
durations_output.pass.cpp:315:18: error: constexpr variable 'integer' must be initialized by a constant expression
  CONSTCD14 auto integer = get_units<char>(std::ratio<123>{});
                 ^         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/howardhinnant/Development/date/include/date/date.h:7823:19: note: modification of object of const-qualified type
      'const char' is not allowed in a constant expression
            p_[i] = x[i];
                  ^
/Users/howardhinnant/Development/date/include/date/date.h:7874:12: note: in call to 'string_literal(string_literal<CharT, 2UL>{x},
      string_literal<CharT, 4UL>{y})'
    return string_literal<CharT, N1 + N2 - 1>{string_literal<CharT, N1>{x},
           ^
/Users/howardhinnant/Development/date/include/date/date.h:7971:28: note: in call to 'operator+(msl(char{'['}), msl<R::num>())'
    return msl(CharT{'['}) + msl<R::num>() + msl(CharT{']'});
                           ^
/Users/howardhinnant/Development/date/include/date/date.h:8186:12: note: in call to 'msl({})'
    return msl<CharT>(p) + string_literal<CharT, 2>{'s'};
           ^
durations_output.pass.cpp:315:28: note: in call to 'get_units({})'
  CONSTCD14 auto integer = get_units<char>(std::ratio<123>{});
                           ^
```

tomaszkam:
Could you please run the test file against the code before my changes in
C++14 mode. The function referenced was not touched by me.

wt., 19.06.2018, 15:35 użytkownik Howard Hinnant <notifications@github.com>
napisał:

HowardHinnant:
I'm not positive how to interpret "the code before my changes in C++14 mode".  If you specify a commit,  I can test it.  Thanks.

HowardHinnant:
Or do you mean against my current master/HEAD?

tomaszkam:
Yes against current master, suffix were meant to be constexpr in C++14
before my changes.

wt., 19.06.2018, 16:33 użytkownik Howard Hinnant <notifications@github.com>
napisał:

HowardHinnant:
I get the same error, and in more places.

HowardHinnant:
Just tried the same experiment with a tip-of-trunk clang, and it passes.  It looks like we're looking at a clang bug in the version of clang I'm using.

What compiler are you using?

tomaszkam:
clang-5.0 and gcc-7 if I remember correctly, will check once I will have
access to my desktop.

What version you are using?



wt., 19.06.2018, 16:43 użytkownik Tomasz Kamiński <tomaszkam@gmail.com>
napisał:

HowardHinnant:
```
Apple LLVM version 9.1.0 (clang-902.0.39.2)
Target: x86_64-apple-darwin17.6.0
Thread model: posix
```
There's no way I know of to map Apple's clang version to a real clang version.

HowardHinnant:
I'm going to see if I can figure out a workaround.  If I can't, I'll go ahead and commit it anyway.

tomaszkam:
Ok, it also does not work with clang++-5.0 with C++14, and works only with g++-7. I must have missed that.

HowardHinnant:
By "do not work", do you mean that common use cases don't work, or that new, stricter tests are failing?  I don't mind if some tests point out compiler bugs, as long as the lib can still work for more common use cases, including with older compilers.

HowardHinnant:
Very nice!

tomaszkam:
I meant that the new constexpr checking test for integral literal was not working also for clang++-5.0, but I missed that accidentally. But this literal was actually not constexpr on master due compiler bug - may test just exposed that.

However, I found a fix - I create an concatenation result inside of operator+, instead of constructor. Made this operator friend to have access to private value_initializing constructor and _p array.

tomaszkam:
Now I am getting errors from gcc, about multiple definitions of operator+, lets me check on that.

HowardHinnant:
Thanks.  I'm running the full test on Apple's clang now...

HowardHinnant:
Confirmed everything still passes here.  Thanks!

tomaszkam:
Tested on Wandbox - in C++14 mode latest version compiles from g++ 5.1 and clang 3.5 (I have not checked earlier versions).

HowardHinnant:
`<chrono>` itself starts falling apart much earlier than that.

tomaszkam:
Tested C++11 for from gcc 4.9 and clang starting from 3.5. Link: https://wandbox.org/permlink/Hir7y2YjYmJzJJAN
I think we may squash it to master now. Hope that there will be no surprises from MSVC.

tomaszkam:
I have restored few inline specifiers on functors, that I have accidentally dropped.

HowardHinnant:
Merged: https://github.com/HowardHinnant/date/commit/2d282e35fa18119d6d9396813da2cc4ad29a9d5a

Thanks much!  Nice job!

PR #357, opened 2018-06-24
Configurable to_local for utc_clock

tomaszkam:
Now the to_local function for utc_clock, throws on default when
time represents leap second insertion. To get a time point before/after
the insertion, the choose argument needs to be provided.

I think that this is step in right direction, unsure only about the exception name.


HowardHinnant:
Closing.  This was a good suggestion, but I have respectfully declined to take this direction.

PR #363, opened 2018-07-16
Make HAS_STRING_VIEW compiler definition part of the public interface

alexanderbock:
Consumers of the library has to be informed about the state of HAS_STRING_VIEW as it changes functions in the public interface.


HowardHinnant:

PR #366, opened 2018-08-02
Export all cmake targets

### LINKED ISSUE #367: CMake export set is incomplete
# Problem
No target for the `tz` library is exported by `CMakeList.txt`.
# Possible solution
* The package configuration exported by `date` should export all targets which may be used by dependent projects.

urp:
Hi,
this fixes  #367 by providing a complete package configuration and also modernizes the `CMakeLists.txt` file:
* Remove target `date_interface`
* Introduce alias and export targets `date::date` and `date::tz`
  * Export a build package configuration into the [cmake registry](https://cmake.org/cmake/help/v3.0/manual/cmake-packages.7.html#package-registry)
  * Export an install package configuration
* Add version information to the project and the exported package configurations
* Minimize use of cmake variables
* Replace old-fashioned directory-level commands (`include_directories`,...) with target-level counterparts (`target_include_directories`,...) 

Note: 
* using targets in namespaces (containing `::`) allows cmake to detect missing dependencies during configuration (instead of linking)
* see this great [talk](https://www.youtube.com/watch?v=bsXLMQ6WgIk) for details on package configurations and modern cmake usage

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -52,8 +52,8 @@ add_library( date::date ALIAS date )
 
 # add include folders to the INTERFACE and targets that consume it
 target_include_directories( date INTERFACE
-  "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>"
-  "$<INSTALL_INTERFACE:$<INSTALL_PREFIX>/include>"
+	"$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>"
+	"$<INSTALL_INTERFACE:$<INSTALL_PREFIX>/include>"
 )
 
 add_library( tz ${HEADER_FILES} ${SOURCE_FOLDER}/tz.cpp )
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -1,9 +1,11 @@
 cmake_minimum_required( VERSION 3.1.0 )
 
-project( date_prj )
+project( date_prj VERSION 2.4.1 )
```
dscole: How about picking up the version directly from git via git-describe?

Basically doing this:

```
-project( date_prj VERSION 2.4.1 )
+execute_process( COMMAND git describe --tags --abbrev=0
+    OUTPUT_VARIABLE version_from_git
+    OUTPUT_STRIP_TRAILING_WHITESPACE )
+
+string( REPLACE "v" "" version_from_git ${version_from_git} )
+
+message( STATUS "Version from git tag: ${version_from_git}" )
+
+project( date_prj VERSION ${version_from_git} )
```

Erroneous1: Would that work if you download from a source tarball?

urp: I see two possible problems:
* It would make the whole build dependent on git if not optional.
* Handling all error cases requires some more boilerplate code

dscole: Yeah, valid concerns and I made some changes to be able to use this locally, but forgot to update my comment here. The way I see it, the way github generates tarballs for source releases forces projects to hard code the version somewhere in the version controlled source. Leaving developers with one of 2 options:
1. Keeping a version hardcoded in the version controlled source code, in addition to creating matching tags.
2. Use a custom tarball packaging script to generate (from git tag) and add some version.txt file to source tree before creating a tarball and loading it to github.
 
In my local checkout I added a `get_version` function that supports 4 different scenarios:
1. User can add a `-DEXPLICIT_VERSION` option to the cmake invocation which will set the project's version.
2. If a version.txt file exists, read it and use the contents as the project's version.
3. Use git describe to produce a version from the latest tag.
4. If all the previous steps failed to produce a valid version, call the `project()` function without a version (the project will be unversioned).

I see no point in even trying to push my change because it requires the maintainer to start generating releases with a special script that also generates the version.txt into the source tarball. 
It would really be great if github were to auto generate such a version.txt file in the standard release mechanism, but not sure how feasible that is.

Just my 2c.

m4ce:
Any chance this could be merged? I am trying to use this library with cmake and it complains that there's no version available when using ```find_package(date 2.4.1)```

HowardHinnant:

PR #372, opened 2018-08-12
add compile definitions as PUBLIC

NeroBurner:
add compile definitions with PUBLIC flag. With this change the project
can be used as subproject in other client projects


HowardHinnant:

PR #373, opened 2018-08-16
Improved C++17 support

Fylax:
Some C++ compiler (such as clang 5.0) have std::void_t and std::string_view in std::experimental namespace.
This workaround checks if the compiler has the experimental namespace and includes the correct files (remapping them to std namespace)

### DIFF
--- a/include/date/date.h
+++ b/include/date/date.h
@@ -69,7 +69,7 @@
 #    elif __has_include(<experimental/string_view>)
 #      include <experimental/string_view>
        namespace std {
-           typedef experimental::string_view string_view;
+           using string_view = experimental::string_view;
        }
 #    endif
 #  endif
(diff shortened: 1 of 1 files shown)


include/date/date.h
```diff
@@ -63,7 +63,16 @@
...
 #include <string>
 #if HAS_STRING_VIEW
-# include <string_view>
+#  if defined(__has_include)
+#    if __has_include(<string_view>)
+#      include <string_view>
+#    elif __has_include(<experimental/string_view>)
+#      include <experimental/string_view>
+       namespace std {
+           typedef experimental::string_view string_view;
```
mdrost: Maybe it would be better to actually use C++17 features like `using string_view = experimental::string_view;` :P

Fylax: Totally agreed. I was mixing C and C++ by the time I wrote the code and I got confused. I fixed it, thank you!

HowardHinnant:
I'd rather not have this library reach into `experimental` at this point.  Developers can do this with `-I<include-path>` and `using` if desired.

PR #374, opened 2018-08-20
In CMake add install EXPORT for tz target

sogartar:
This fixes the problem of having a target that
depends on tz and it has "install(EXPORT ...)".

Without install EXPORT for tz you would get:
CMake Error: INSTALL(EXPORT "mytarget" ...) includes target "mytarget" which requires target "tz" that is not in the export set.


sogartar:
Just to clarify. This happens when you use `subdirectory_add` in CMake to bring in the `date` project and having another target that depends on `tz`, that has INSTALL EXPORT of a CMake config.

HowardHinnant:

PR #376, opened 2018-08-28
Detect std::string_view using __cpp_lib_string_view.

mbeach:
This allows us to cope with situations where the compiler may support
C++17, but the standard library implementation may not be complete.

I ran into this situation when attempting to build with clang and libc++ on Ubuntu 17.10, which has clang 5.0 but ships with libc++ 3.9.1. The compiler sets __cplusplus to a value which indicates C++17 support, but the library does not have std::string_view.

### DIFF
--- a/include/date/date.h
+++ b/include/date/date.h
@@ -31,14 +31,6 @@
 // been invented (that would involve another several millennia of evolution).
 // We did not mean to shout.
 
-#ifndef HAS_STRING_VIEW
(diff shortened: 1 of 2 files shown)


HowardHinnant:
The intent is that the developer can control this by defining `HAS_STRING_VIEW`.  This PR only changes the default for when `HAS_STRING_VIEW` kicks in automatically.  I'd rather not rock the boat on this at this time.

PR #379, opened 2018-09-03
update CXX_STANDARD setting

NeroBurner:
Set minimum CXX_STANDARD depending if string_view is enabled or not.
Don't override CMAKE_CXX_STANDARD variable if it is set

When I used tz as a submodule I set string_view usage to disabled, but CMake still said tz required C++17. With this change, the CMAKE_CXX_STANDARD variable is not modified when it is set before (by a user or by the parent CMake project using tz project

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,9 +6,6 @@ include( GNUInstallDirs )
 
 find_package( Threads REQUIRED )
 
-# Override by setting on CMake command line.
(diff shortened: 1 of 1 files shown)


NeroBurner:
rebased commit on current master

NeroBurner:
@HowardHinnant rebased on current master

HowardHinnant:

Issue #368, opened 2018-08-03
Provide the database as an input stream


kivadiu:
I would like to embed the database into my program at compile time. If I cat all required files and remove all comments, its only 156 kb which can be less than 25 kb when compressed.
```
#!/bin/bash
(
  for f in africa antarctica asia australasia backward etcetera europe pacificnew northamerica southamerica systemv leapseconds; do
    cat $f
  done
) | grep -v "^#.*" | grep -v "^$"
```

For this to work, I would need an option to load the database from a `std::istream`.

What do you think of that? I am ready to make a PR if you give me some ideas of what's the best to modify.

Regards

HowardHinnant:
Your requirement is a bit ambiguous:  `std::istream` can't be used at compile-time.  You can only use it to load data at run-time.

But I get that you would like an alternative tzdb.  You can get part way there just by using the uncompressed-but-comment-free form as your database.  Just do that part and point INSTALL toward wherever you put it, and you're good to go, except you'll need the version file too (just a few more bytes).

To go further than that, you might look at the possibility of providing a user-defined custom database.  If you look at [Custom time zone](https://howardhinnant.github.io/date/tz.html#Examples), there is a tiny example of how to insert your own timezone into this library.

(Here)[https://github.com/HowardHinnant/date/blob/master/include/date/ptz.h] is a much more complex example that models the complete POSIX timezone spec.  Neither of these does exactly what you want.  However they do demonstrate that one can do a great deal of customization in this area.

kivadiu:
I expressed my-self badly. What I meant by compile-time was that I could include the whole database in a binary array. And then at runtime, I can read it with std::istream. What I would like is no file at all, everything in the binary.

HowardHinnant:
Another option that is already available is compiling with  `-DUSE_OS_TZDB`.  With this option the "zic-compiled" form of the tzdb is read from the files.  On non-Windows platforms, the OS typically supplies this for you, though there is nothing stopping you from compiling your own and using it.

In this form, the data is a binary-formatted list of transition points for each time zone.  It typically loads more quickly.  On macOS, the Apple-supplied files take up about 94Kb on disk.  I'm not sure of the size on linux systems.

You can learn more about using the zic compiler at https://www.iana.org/time-zones

At this time I am not prepared to support a 3rd format for the tzdb.

PR #385, opened 2018-09-14
Standalone version allowing to provide the database as std::istream

kivadiu:
This makes it possible to embed the database in a string in the source
code so that the library is completely standalone. No need to download
the database, no need to read it from disk files.

Notice that this not a new format. The embedded file should just be a concatenation of the IANA database files.

With this commit, you can compile with -DHAS_REMOTE_API=0
-DAUTO_DOWNLOAD=0 and still have access to the database like this:

date::set_iana_db_source_stream(&iana_getter);
date::set_windows_xml_source_stream(&win_xml_getter);

where iana_getter() and win_xml_getter() both return a
std::unique_ptr<std::istream>.

The stream returned by iana_getter() is the concatenation of files:
version africa antarctica asia australasia backward etcetera europe
pacificnew northamerica southamerica systemv leapseconds from
https://data.iana.org/time-zones/releases/tzdata${version}.tar.gz.

The stream returned by win_xml_getter is the content of the file
http://unicode.org/repos/cldr/trunk/common/supplemental/windowsZones.xml.

Fixes issue #368.

Example of how it can be used in the attached file.
[standalone-db.cpp.gz](https://github.com/HowardHinnant/date/files/2382396/standalone-db.cpp.gz)

### DIFF
--- a/include/date/tz.h
+++ b/include/date/tz.h
@@ -2757,6 +2757,18 @@ to_gps_time(const tai_time<Duration>& t)
 
 #endif  // !MISSING_LEAP_SECONDS
 
+using database_stream_getter = std::unique_ptr<std::istream> (*)(void);
+
+// set IANA database source as intput stream
+DATE_API void
+set_iana_db_source_stream(database_stream_getter);
+
+#ifdef _WIN32
+// set Windows time zone database source as input stream
+DATE_API void
+set_windows_xml_source_stream(database_stream_getter);
+#endif
+
 }  // namespace date
 
 #endif  // TZ_H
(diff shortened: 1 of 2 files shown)


kivadiu:
Hi Howard,
Did you get a chance to look at this PR?
Kind regards,
F

HowardHinnant:
I have not had a chance yet, but I haven't forgotten about it.  Sorry for the delay.

kivadiu:
Thank you, take your time.

HowardHinnant:
After much thought, this feature is not something I want to have the responsibility of supporting in the future.  I do think it is a neat idea, and have no objection to you supporting it on your fork.

My rationale for not wanting to support it myself is the knowledge that the IANA database gets updated on average about every 2 months, so I don't want to lure the majority of my clients into locking the database into their binary.  I fear I would get too many complaints on this feature from users who don't fully understand it.  And within a couple of years I'll be advising my clients to adopt C++20 in lieu of this library, which can't support this feature anyway.

PR #386, opened 2018-09-14
Prefer using std::string that using namespace std and unqualified string

kivadiu:
'using namespace std;' in header files can conflict with custom
namespace names. For example 'string' is used without 'std::'
qualification in tz.h. If tz.h is included in a file where string is a
namespace, it cannot be compiled anymore.

This is the case in my program where I put anything related to strings inside namespace string.

This fixes only the issue with string but I think other using namespace should be removed from all .h to avoid any bad surprises.

### DIFF
--- a/include/date/tz.h
+++ b/include/date/tz.h
@@ -1963,10 +1963,9 @@ std::basic_ostream<CharT, Traits>&
 to_stream(std::basic_ostream<CharT, Traits>& os, const CharT* fmt,
           const utc_time<Duration>& t)
 {
-    using namespace std;
     using namespace std::chrono;
-    using CT = typename common_type<Duration, seconds>::type;
-    const string abbrev("UTC");
+    using CT = typename std::common_type<Duration, seconds>::type;
+    const std::string abbrev("UTC");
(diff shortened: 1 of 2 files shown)


include/date/tz.h
```diff
@@ -1966,7 +1966,7 @@ to_stream(std::basic_ostream<CharT, Traits>& os, const CharT* fmt,
     using namespace std;
```
tweej: I agree with you, but your change doesn't remove the using directive here.

kivadiu: I left it because of common_type but if you agree, I can update the PR to remove it.
In my opinion, it would also be better to remove using std::chrono.

kivadiu:
To illustrate the problem, this program does not compile:
```
#include <string>

namespace string {
struct foo {};
} // namespace string

std::string bar() {
  using namespace std;
  return string{};
}

int main() {
  bar();
  return 0;
}
```
And this is what happens when I include `tz.h` but I see that it does the same if I include `date.h` with ONLY_C_LOCALE defined to 1. So let's make another modification for that.

Issue #387, opened 2018-09-14
parse error does not lead to ok() == false


kivadiu:
It was a surprise to me that parsing a bad date did not always lead to `ok()` returning `false`. This simple program:
```
std::istringstream iss{"201800"};
date::year_month ym;
iss >> date::parse("%4Y%2m", ym);
std::cout << "ym.ok() == " << ym.ok() << '\n';
```
leads to undefined behaviour (confirmed by valgrind).

This is because `year`, `month` and `day` objects are not initialized by default and because `from_stream` does not set `ok` (probably because `ok` is not stored in the object, it's a function, and `from_stream` does not know how to set an object to an invalid state).

After reading the source code, I discovered that `from_stream` sets the stream to a failed state so the correct method is to check the state of the stream after parsing, not the value of the date object. This is consistent with the behaviour of `>>` operator with standard types so it is perfectly fine. However, I would suggest to document the fact that an uninitialized object is in an undefined state (could be ok or not) and that failure to parse will not change its value. Also that calling ok() on an uninitialized object is undefined behaviour.

I do not find the documentation in the repository so I am sorry not to be able to propose a PR for that.

HowardHinnant:
The documentation is on the https://github.com/HowardHinnant/date/tree/gh-pages branch.  For this particular function in:  https://github.com/HowardHinnant/date/blob/gh-pages/date.html

`year_month` says:
```
class year_month
{
    date::year  y_;  // exposition only
    date::month m_;  // exposition only

public:
    year_month() = default;
...
```
which is enough (at least recursively) to specify the "`int`-like" default construction behavior.  I.e. default initialization is unspecfied, but zero-initialization can also be used: `year_month ym{};`.

I agree that `from_stream` should be better specified to not alter the target parsed object if there is a parse failure.

kivadiu:
> I agree that `from_stream` should be better specified to not alter the target parsed object if there is a parse failure.

Please find a proposed addition here: #389.

PR #389, opened 2018-09-16
Better explain that from_stream does not change the value of the target

kivadiu:
Better explain that `from_stream` does not change the value of the target in case of an error and that `!ok()` is not an indication of error.

Fixes issue #387.


kivadiu:
Thanks for merging.

PR #394, opened 2018-09-19
Define for std::uncaught_exceptions(). 

molind:
Xcode 10 uses clang and perfectly compiles C++17, but std::uncaught_exceptions() is available only on iOS 10+ devices. When application supports iOS 9 devices and uses C++17 there is no way to build it, except drop iOS 9 support or remove C++17 code. Using this define we could configure should date.h use std::uncaught_exceptions() or not.

### DIFF
--- a/include/date/date.h
+++ b/include/date/date.h
@@ -132,7 +132,7 @@ namespace date
 #  else
 #    define HAS_UNCAUGHT_EXCEPTIONS 0
 #  endif
-#endif  // HAS_VOID_T   
+#endif  // HAS_UNCAUGHT_EXCEPTIONS   
 
 #ifndef HAS_VOID_T
 #  if __cplusplus >= 201703
--- a/include/date/date.h
+++ b/include/date/date.h
@@ -126,14 +126,6 @@ namespace date
 #  define NOEXCEPT noexcept
 #endif
 
-#ifndef HAS_UNCAUGHT_EXCEPTIONS
-#  if __cpp_lib_uncaught_exceptions
-#    define HAS_UNCAUGHT_EXCEPTIONS 1
-#  else
-#    define HAS_UNCAUGHT_EXCEPTIONS 0
-#  endif
-#endif  // HAS_UNCAUGHT_EXCEPTIONS   
-
 #ifndef HAS_VOID_T
 #  if __cplusplus >= 201703
 #    define HAS_VOID_T 1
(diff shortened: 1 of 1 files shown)


beached:
Why not use the feature test macro __cpp_lib_uncaught_exceptions

molind:
@beached Didn't know about it. :) Updated detection to `__cpp_lib_uncaught_exceptions`.

molind:
May be whole define construction is redundant if we could detect it precisely in place. What do you think?

molind:
BTW, there is __cpp_lib_void_t test macro.

https://github.com/HowardHinnant/date/blob/ad6a7c95644e7248d653e227e1f0dac9572d2851/include/date/date.h#L137-L143

May be it's also should be detected in place.

HowardHinnant:
At this point do we need `HAS_UNCAUGHT_EXCEPTIONS` at all.  Why not just use `__cpp_lib_uncaught_exceptions` directly in the one place where we need it?  Or do you anticipate the need to override the default behavior of `HAS_UNCAUGHT_EXCEPTIONS == __cpp_lib_uncaught_exceptions`?

molind:
@HowardHinnant Ok, I removed `HAS_UNCAUGHT_EXCEPTIONS`.

beached:
Can't one assume that the feature test macro's are there and if not the feature is unsupported?

molind:
I checked it again on Xcode 10.0
Expected behavior: `__cpp_lib_uncaught_exceptions` macro should depend on Deployment Target setting and selected C++ Language Dialect. 
Actual results: `__cpp_lib_uncaught_exceptions` is undefined. I've tried deployment iOS 9 - 12 deployment targets, libc++ standard library and C++17. However `std::uncaught_exceptions()` works without issues since deployment target iOS 10.0.

beached:
ooh this is the same as #416.  Why not just use the feature flag?  It may work but so will uncaught_exception until c++20.  command line wth Mojave/Apple LLVM version 10.0.0 (clang-1000.11.45.5) doesn't set the flag either.

HowardHinnant:
Sorry for the long delay.

In https://github.com/HowardHinnant/date/issues/416 Marshall made a good argument for not using `__cpp_lib_uncaught_exceptions`.  However I like the flexibility that `HAS_UNCAUGHT_EXCEPTIONS` gives the developer in this P/R.  I've adopted this, fixed the type-o's., and changed the test default to turn this option on at `__cplusplus > 201703` because `std::uncaught_exception()` is removed in the draft C++20 spec.

Thank you for your patience.  Merged:  https://github.com/HowardHinnant/date/commit/5ba1c1ad8514362dba596f228eb20eb13f63d948

PR #401, opened 2018-11-01
Fix bug parsing locale date (%x)

AdrienBellanger:
Parsing istringstream formated as "%x" and "%x %X" was switching day and month.

### DIFF
--- a/include/date/date.h
+++ b/include/date/date.h
@@ -6376,7 +6376,7 @@ from_stream(std::basic_istream<CharT, Traits>& is, const CharT* fmt,
                     {
 #if !ONLY_C_LOCALE
                         ios_base::iostate err = ios_base::goodbit;
-                        f.get(is, nullptr, is, err, &tm, command, fmt+1);
+						f.get_date(is, nullptr, is, err, &tm);
                         if ((err & ios::failbit) == 0)
                         {
                             checked_set(Y, tm.tm_year + 1900, not_a_year, is);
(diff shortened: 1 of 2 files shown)


HowardHinnant:
Sorry, there's several things I can't accept with this PR.

*  This looks like it is working around a bug in the `std::time_get` facet of an implementation + platform.  The test _almost_ works on my platform, without the fix.  This wouldn't be the first `time_get` bug discovered.

*  The locale names in the test aren't portable.  On my platform neither locale is found.  Instead I have "en_US" and "de_DE".

*  On my platform "en_US" is looking for mm/dd/yyyy, not mm/dd/yy.  So the test gets the year 18 instead of the year 2018.

*  The `istringstream` isn't cleared or reset in any way after the first parse, so the second parse fails no matter what.

*  On my platform "de_DE" is looking for dd.mm.yyyy, so the second test fails to parse.

I don't know of a good way of testing locale-sensitive parsing since locales are unfortunately so non-portable.

This test passes for me, without your patch, but probably won't for you:

```
#include "date/date.h"
#include <cassert>
#include <iostream>
#include <type_traits>
#include <locale>

int
main()
{
    using namespace date;
    using namespace std::chrono;
    std::locale loc("en_US");
    std::istringstream iss("3/10/2018");
    auto pos = iss.tellg();
    iss.imbue(loc);
    date::local_seconds tLocal;
    iss >> date::parse("%x", tLocal);
    // Seconds since Epoch for the 10th of march
    assert(tLocal.time_since_epoch().count() == 1520640000);
    iss.imbue(std::locale("de_DE"));
    iss.clear();
    iss.str("03.10.2018");
    iss >> date::parse("%x", tLocal);
    // Seconds since Epoch for the 3rd of october
    assert(tLocal.time_since_epoch().count() == 1538524800);
}
```

Issue #331, opened 2018-04-13
Issue with MSVC and /Zc:__cplusplus


AndrewPardoe:
The `#if` side of the definition of `string_literal` on line 7522 fails with a correctly set `__cplusplus` in MSVC 15.7 preview 3 and above. The `#else` side on 7908 appears to work. 

We know about the bug (#600264) and will fix it.

HowardHinnant:
Thanks for letting us know!  If you decide you need me to shuffle MSVC back to the `#else` side, just let me know.  But I'm glad this little project is helping make MSVC better and better.

AndrewPardoe:
Hopefully we can fix this quickly. If not, we might ask for a one-off.

xiangfan-ms:
BTW, the two source workarounds:
1. changing 'std::enable_if_t<1 < sizeof(U)>' to 'std::enable_if_t<(1 < sizeof(U))>'
 While the parentheses aren't required, it does clarity the source code and improves the readability.
2. changing 'std::is_same<>{}' to 'std::is_same_v<>')
 I think the workaround is actually better than the current implementation :-). It is semantically the same and is cheaper to compile.

HowardHinnant:
I haven't reviewed these changes in detail yet.  But I definitely want VS-latest to work with this lib in as high a quality as practical.  So I'm highly motivated.  Things I'll be looking for are C++11 compatibility.  But I suspect we're talking about a C++14 and higher branch anyway.  And I have a previous history of making minor syntactical changes to accommodate other compilers.

This is my night job and I took last night off. :-)  If you put up a P/R, I'll definitely review it, and if I don't accept it, I'll recommend changes you can make to get it accepted.

Stylistically, I prefer  `std::is_same<>{}` to `std::is_same_v<>`.  I wasn't crazy about the `_v` additions for that reason, though I did not object to them either.  They introduced an alternative name for something that already had a good name with the same number of characters.  Its only advantage was to benefit platforms that were behind in `constexpr` compliance.  But style is a secondary concern to getting all platforms up and running.  I mention it just because you asked.

HowardHinnant:
Oh, just noticed that `is_same_v` is C++17.  No can do.  I can do `is_same<>::value` instead.  Objectively uglier style, but up-and-running takes precedence.

HowardHinnant:
Not resolved yet.  I need someone besides me to take point on VS.  I can advise, but I can't test.

zlojvavan:
[...]

PR #410, opened 2018-12-16
MSVC accepts zoned_time(std::string_view name, ...) constructors now.

DanielaE:
With '/Zc:__cplusplus' applied, MSVC also enables std::string_view, causing compilation errors in zoned_time constructors which take string_view arguments. The reason for this behaviour is the template argument 'T' in the declaration of these constructors which are missing the definition later in the file. Adding them makes MSVC happy again.

Signed-off-by: Daniela Engert <dani@ngrt.de>

### DIFF
--- a/include/date/tz.h
+++ b/include/date/tz.h
@@ -343,13 +343,13 @@ private:
     sys_time<duration> tp_;
 
 public:
-#if !defined(_MSC_VER) || (_MSC_VER > 1915)
+#if !defined(_MSC_VER) || (_MSC_VER > 1900)
     template <class T = TimeZonePtr,
               class = decltype(zoned_traits<T>::default_zone())>
 #endif
         zoned_time();
 
-#if !defined(_MSC_VER) || (_MSC_VER > 1915)
+#if !defined(_MSC_VER) || (_MSC_VER > 1900)
     template <class T = TimeZonePtr,
               class = decltype(zoned_traits<T>::default_zone())>
 #endif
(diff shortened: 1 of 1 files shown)


DanielaE:
@HowardHinnant 
Howard, what do you think about this solution to deal with the VS2017 issue #331? It feels much cleaner imho. It compiles and runs just fine with my Visual Studio test project using both the latest msvc compiler (19.16 from VS2017 update 9, earlier 19.1x are fine as well) and clang 8.0 (I have no earlier ones available in Visual Studio).

If you happen to like it, I'll rebase it on HEAD. Is there some CI in place to precheck a PR?

HowardHinnant:
I'm going to have to file a bug report against github for the missing face-palm emoji. ;-)

Yes, if this works I like it much better.  Thanks much for looking into this!

There is no CI to precheck a PR.  I've looked into CI before and prefer not to use it.

DanielaE:
Nevermind, I'm no better :smile: : I've had this laying around for months now, just being too lazy to fork your repo and open a PR. Take it as a christmas 🎁  from me.

HowardHinnant:
Done, thanks. https://github.com/HowardHinnant/date/commit/7231a182a4eed98a2766472cd3f364105d7ba6fb

DanielaE:
@HowardHinnant: is there  reason why you left out dddec56?

HowardHinnant:
I squashed your commits together in an attempt to simplify the history.  Those changes are on master now.  Admittedly I probably could have picked a better commit message.

DanielaE:
Sorry, I must have been blind - totally right you are. If I had known that you are in the "pre-squashed" camp, I would have done the history-edit myself. Thx, Dani

PR #411, opened 2018-12-18
Ignore tzdata.zi and leapseconds tzdb files on Linux when loading time zones

apretori-tic:
Hi Howard

Thanks for this great library!
When loading tzdb files on Linux 2 additional files are considered to be time zones.  And when enumerating time zones they are listed as valid time zones. This patch just skips them.

Kind regards,
Albert


HowardHinnant:
Thanks!

PR #430, opened 2019-03-20
allow to build shared or static library with position independent code support (fPIC)

m4ce:
I have the need to build both static and shared libraries with fPIC support.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -45,7 +45,8 @@ set( HEADER_FILES
 	${HEADER_FOLDER}/date/tz_private.h
 )
 
-add_library( tz ${HEADER_FILES} ${SOURCE_FOLDER}/tz.cpp )
+add_library(tz ${HEADER_FILES} ${SOURCE_FOLDER}/tz.cpp )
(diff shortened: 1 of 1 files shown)


jkhoogland:
I have the same requirement.
Would be great to have this merged.

CMakeLists.txt
```diff
@@ -45,7 +45,12 @@ set( HEADER_FILES
 	${HEADER_FOLDER}/date/tz_private.h
 )
 
-add_library( tz ${HEADER_FILES} ${SOURCE_FOLDER}/tz.cpp )
+if (BUILD_SHARED_LIBS)
```
NeroBurner: no need for this if-else block, `add_library` recognizes `BUILD_SHARED_LIBS`

> If no type is given explicitly the type is STATIC or SHARED based on whether the current value of the variable BUILD_SHARED_LIBS is ON

https://cmake.org/cmake/help/v3.14/command/add_library.html?highlight=add_library

m4ce: Thanks - amended.

HowardHinnant:

PR #451, opened 2019-05-20
#endf -> #endif after recent change

joncppl:
Thanks for fixing the depreciation warning. 
Looks like there is a typo though ~~


HowardHinnant:
Thanks!  The one time I think a change is too trivial to run tests... ;-)

PR #454, opened 2019-05-22
Use realpath the determine the target of the localtime symlink

Hubble1942:
The previous solution would fail, if `/etc/localtime` is a symlink to a symlink to the zone details. Using realpath resolves all potential symlinks and returns directly the path of interest.

### DIFF
--- a/src/tz.cpp
+++ b/src/tz.cpp
@@ -3694,14 +3694,18 @@ tzdb::current_zone() const
             
             auto rp = realpath(timezone, nullptr);
             if (rp == nullptr)
-                throw system_error(errno, system_category(), "realpath() failed");
+                throw system_error(errno, system_category(), R"(realpath("/etc/localtime") failed)");
             
             auto result = string(rp);
             free(rp);
 
-            const size_t pos = result.find(get_tz_dir());
+            const string foldername{"zoneinfo"};
+            const size_t pos = result.find(foldername);
             if (pos != result.npos)
-                result.erase(0, get_tz_dir().size() + 1 + pos);
+                result.erase(0, foldername.size() + 1 + pos);
+            else
+                throw system_error(errno, system_category(), R"(failed to find "zoneinfo-part in path to local timezone file")");
+
             return locate_zone(result);
         }
     }
--- a/src/tz.cpp
+++ b/src/tz.cpp
@@ -3691,12 +3691,13 @@ tzdb::current_zone() const
         CONSTDATA auto timezone = "/etc/localtime";
         if (lstat(timezone, &sb) == 0 && S_ISLNK(sb.st_mode) && sb.st_size > 0) {
             using namespace std;
-            string result;
-            char rp[PATH_MAX+1] = {};
-            if (readlink(timezone, rp, sizeof(rp)-1) > 0)
-                result = string(rp);
-            else
-                throw system_error(errno, system_category(), "readlink() failed");
+            
+            auto rp = realpath(timezone, nullptr);
+            if (rp == nullptr)
+                throw system_error(errno, system_category(), "realpath() failed");
+            
+            auto result = string(rp);
+            free(rp);
 
             const size_t pos = result.find(get_tz_dir());
             if (pos != result.npos)
--- a/src/tz.cpp
+++ b/src/tz.cpp
@@ -3700,7 +3700,7 @@ tzdb::current_zone() const
             free(rp);
 
             const string foldername{"zoneinfo"};
-            const size_t pos = result.find(foldername);
+            const size_t pos = result.rfind(foldername);
             if (pos != result.npos)
                 result.erase(0, foldername.size() + 1 + pos);
             else
--- a/src/tz.cpp
+++ b/src/tz.cpp
@@ -3704,7 +3704,7 @@ tzdb::current_zone() const
             if (pos != result.npos)
                 result.erase(0, foldername.size() + 1 + pos);
             else
-                throw system_error(errno, system_category(), R"(failed to find "zoneinfo-part in path to local timezone file")");
+                throw system_error(errno, system_category(), R"(failed to find "zoneinfo"-part in path to local timezone file)");
 
             return locate_zone(result);
         }


HowardHinnant:
Have you experienced this failure mode (symlink to a symlink)?

Hubble1942:
Yes, I have. Its on an embedded system, where /etc is on a read-only mounted partition. To enable switching the timezone, we have a symlink to a writable partition and from there a symlink to the final timezone file.

HowardHinnant:
When running this program on macOS 10.14.5 (Mojave):
```
#include "date/tz.h"
#include <iostream>

int
main()
{
    using namespace date;
    using namespace std::chrono;
    std::cout << zoned_time{current_zone(), system_clock::now()} << '\n';
}
```
I get this run-time error:
```
libc++abi.dylib: terminating with uncaught exception of type std::runtime_error: /private/var/db/timezone/tz/2019a.1.0/zoneinfo/America/New_York not found in timezone database
```
It seems the string "/private/var/db/timezone/tz/2019a.1.0/zoneinfo/America/New_York" is being passed to `locate_zone`, and this function is expecting only "America/New_York".

Hubble1942:
As I do not have access to a mac, I cannot reproduce this issue.

But it looks like, my change conflicts with the `__APPLE__` specific implementation of `discover_tz_dir()`:

By design, `realpath` returns an **absolute** path. In your case `/private/var/db/timezone/tz/2019a.1.0/zoneinfo/America/New_York`.
On mac, `discover_tz_dir()` may return a **relative** path. I guess, something like `../var/db/timezone/tz/2019a.1.0/zoneinfo/America/New_York`.

With this difference, the `result.find(get_tz_dir())` on line 3702 will return `npos` and the full absolute path will be passed to `locate_zone()`.

To solve this issue, I see the following two options:

1. Conditional compilation in `current_zone()`: If `__APPLE__` is defined, use the old implementation using `readlink`. Use my implementation using `realpath` otherwise.
2. In `discover_tz_dir()`, use `realpath` instead of `readlink` as well.

Which solution would you prefer?

HowardHinnant:
`discover_tz_dir()` returns the absolute path to the `zoneinfo` directory, including `zoneinfo`.  On macOS this is currently `/var/db/timezone/zoneinfo`.

On line 3702 `result = /var/db/timezone/zoneinfo/America/New_York`.

With your change, `result = /private/var/db/timezone/tz/2019b.1.0/zoneinfo/America/New_York`.

It looks like Apple has set up `/var` as a symlink to `/private/var`.

Under `/private/var/db/timezone` exists one directory, and three symlinks to directories:

```
Jade:/private/var/db/timezone> ls -la
total 0
dr-xr-xr-x@   6 root  wheel   192 Jul 23 12:01 .
drwxr-xr-x  102 root  wheel  3264 Aug 31 03:18 ..
lrwxr-xr-x    1 root  wheel    35 Jul 23 11:59 icutz -> /var/db/timezone/tz/2019b.1.0/icutz
drwxr-xr-x    3 root  wheel    96 Jul 23 12:01 tz
lrwxr-xr-x    1 root  wheel    29 Jul 16 23:10 tz_latest -> /var/db/timezone/tz/2019b.1.0
lrwxr-xr-x    1 root  wheel    38 Jul 23 11:59 zoneinfo -> /var/db/timezone/tz/2019b.1.0/zoneinfo
```

I'm _guessing_ that this is all being set up in preparation for delivering multiple versions simultaneously of the tzdb in the future.  Perhaps to be able to update the tzdb without requiring a system reboot (just guessing!).

I'm beginning to think that we both have symlinks to symlinks issues, just with slightly different designs by the OS.  And that these symlinks will actually be helpful during a run-time update by the OS of a tzdb (which I have no way of currently testing).

_All_ `current_zone()` is cares about is the trailing "America/New_York" under "/etc/localtime".  Perhaps the smart thing to do would be to continue using `readlink` (keep the OS's indirection), but reverse search the result "zoneinfo", and then pick the trailing information off after that.

I can try to work up such a solution.  But I'm wondering if the reverse search for "zoneinfo" will work for you as well.

Hubble1942:
How about just seeking the `zoneinfo`-part instead of the whole result of `get_tz_dir()`? Would be a similar solution to: https://unix.stackexchange.com/a/451925

I will update the PR accordingly...

HowardHinnant:
My rationale for searching _backwards_ for "zoneinfo" is so that the next person in line doesn't report that his platform has a path such as "/usr/share/db/zoneinfos/2019b/zoneinfo/America/New_York".

Hubble1942:
OIC. You're absolutely right. `rfind` then...

HowardHinnant:
Could you tell me what `readlink(timezone, rp, sizeof(rp)-1) > 0)` fills into `rp` on your system?

HowardHinnant:
How does this work for you? https://github.com/HowardHinnant/date/commit/d54bb6daa640148974f3162d89a5ab2c64ef819f

HowardHinnant:
@Hubble1942 ping.

Hubble1942:
Sorry for my late answer. This is somehow a side project of mine and I had higher priorities recently.

> Could you tell me what `readlink(timezone, rp, sizeof(rp)-1) > 0)` fills into `rp` on your system?

As I expected: `/opt/etc/localtime`. This is the primary target of the `/etc/localtime`-symlink and again, a symlink pointing to the selected timezone in `/usr/share/zoneinfo`.

> How does this work for you? [d54bb6d](https://github.com/HowardHinnant/date/commit/d54bb6daa640148974f3162d89a5ab2c64ef819f)

Unfortunately doesn't work either. With still the same issue.

Is there a reason for not using `realpath()`. In my opinion, as we are only interested in the part after `/zoneinfo/`, it shouldn't make any difference if we resolve all symlinks to the zoneinfo's location. Or am I missing an important point?

HowardHinnant:
Ok, try number next: https://github.com/HowardHinnant/date/commit/d54bb6daa640148974f3162d89a5ab2c64ef819f

This is very similar to your proposal but with the following differences:

* Ask `realpath` to put answer in stack-allocated memory instead of on the heap.  This avoids an exception safety issue here:  https://github.com/HowardHinnant/date/pull/454/commits/c3cb14c10273b79b0480cd21866d79cf94d0791f#diff-b6db5e3c83c32a0f14d8870ca15a7b8fR3699

* Introduce option to use `string_view` if available to eliminate all heap allocations on the non-exceptional path.

* Change exception message to not use `errno` which `rfind` doesn't set.

Does this work for you?

HowardHinnant:
Sorry, my first shot at this message linked to the wrong commit.

Ok, try number next: https://github.com/HowardHinnant/date/commit/cccda506313b110c964f8e0ef37be40458bacc8e

This is very similar to your proposal but with the following differences:

Ask realpath to put answer in stack-allocated memory instead of on the heap. This avoids an exception safety issue here: c3cb14c#diff-b6db5e3c83c32a0f14d8870ca15a7b8fR3699

Introduce option to use string_view if available to eliminate all heap allocations on the non-exceptional path.

Change exception message to not use errno which rfind doesn't set.

Does this work for you?

Hubble1942:
> Ok, try number next: [cccda50](https://github.com/HowardHinnant/date/commit/cccda506313b110c964f8e0ef37be40458bacc8e)

Excellent, works as one would expect. Thank you very much!

I'll withdraw my PR then.

psimon-prnj:
I have observed a recurrence of 252 "timezone detection fails for US/Eastern" under ubunutu 18.04 and g++ 7.4 on two different hardware platforms. The failure traces back to commit 4481 for this issue.

Replacing calls to realpath() with readlink() locally corrects the regression on both systems, but would presumably reopen 454 for affected platforms.

config details: I am working with a program equivalent to the simple test program in the comments above, and compiling with options -DINSTALL="/path/to/date" with tzdata installed there, and HAS_REMOTE_API=0

note: also tested under windows using msys2/mingw64, and macos - neither system impacted by using tz.cpp as is or replacing realpath() with readlink().

HowardHinnant:
If an exception is thrown, could you print out and report the `.what()` of that exception?  Thanks.

psimon-prnj:
here's the full program output:
terminate called after throwing an instance of 'std::runtime_error'
  what():  posixrules not found in timezone database
Aborted (core dumped)

psimon-prnj:
I have confirmed the exception and terminate are happening inside realpath()

HowardHinnant:
Could you print out `result` (or `rp`) in the changed code, both using `realpath`, and `readlink`?  Thanks.

psimon-prnj:
this code change:

            // if (realpath(timezone, rp) == nullptr)
            //     throw system_error(errno, system_category(), "realpath() failed");
            if (readlink(timezone, rp, sizeof(rp)-1) <= 0)
                 throw system_error(errno, system_category(), "readlink() failed");
            std::cout << "readlink() returned with rp = " << rp << std::endl;

generates this output:
readlink() returned with rp = /usr/share/zoneinfo/America/New_York

HowardHinnant:
Thanks, and with `realpath()`?

psimon-prnj:
/usr/share/zoneinfo/posixrules

HowardHinnant:
Wow, thanks.

psimon-prnj:
sorry for the delay on that, I thought a previous test had shown that the exception was inside realpath, but it's not ... investigating

HowardHinnant:
That's ok.  I know it is coming out of `locate_zone`.

HowardHinnant:
I think that to fix this I'm going to need to #ifdef on ubunutu.  Do you know of a macro that says "Hey, this is ubunutu!"?

psimon-prnj:
not offhand ... poking around for one

psimon-prnj:
I'm puzzled by something, /usr/share/zoneinfo/America/New_York is Type: Link to Binary (application/octet-stream) with Link target: ../posixrules

mellery451:
@HowardHinnant macros will only detect the compiling system...don't you need a runtime check for this problem?

HowardHinnant:
I'd settle for a runtime check if I knew how to check for it.

HowardHinnant:
What is `ls -l /usr/share/zoneinfo/posixrules`?

psimon-prnj:
-rw-r--r-- 1 root root 3545 Jul  4  2019 /usr/share/zoneinfo/posixrules

HowardHinnant:
Your help is much appreciated in figuring this out @psimon-prnj!

What do we think about adding a `USE_READLINK` macro that defaults to 0?  For this platform you would have to compile with `-DUSE_READLINK`.

HowardHinnant:
I think I understand what is going on.  `/usr/share/zoneinfo/posixrules` is a copy of the `America/New_York` time zone file.  When you change time zones on this machine it copies that time zone to `/usr/share/zoneinfo/posixrules`.

Unfortunately the contents of the file don't include the name `America/New_York`.  The only way to get the name is to extract it from the symbolic link.

psimon-prnj:
here's a possible run time solution:
check the path returned by realpath()
if it's one the code should ignore (e.g. in the list currently in init_tzdb()) then
call readlink()
check the path returned by that function also, because if it's one we should ignore, the result will be invalid

HowardHinnant:
I could do that.  Although the performance of `current_zone()`  is far more important than the performance of `init_tzdb()`.  `current_zone()` is called many times, and `init_tzdb()` is called only once.  If we go this route, I recommend a separate list of names, starting with a list of one:  "posixrules".  The more forbidden names we add, the greater the runtime cost.

HowardHinnant:
Maybe I could do the test once in `init_tzdb()` and store the result in a `bool` that `current_zone()` tests.

psimon-prnj:
I think you just said something very similar to what I was typing, but just to be sure ...

One way around the efficiency issue would be a three-value static flag, with values for uninitialized, use_realpath, and use_readlink. That provides the freedom to make the list as long as necessary, or alternately, to just check the first time if the first call attempted returned a valid time zone name. (This should handle most use cases, but could be problematic for an always-on system.)

I also don't know which call should be first, e.g. call readlink() first, check if the result is a valid time zone name, then call realpath() (first time through only). On the face of it, I would expect the use cases covered by multiple symlinks (this issue 454) to be less common than the single symlink case (prior issue 252), but that intuition could be wrong.

HowardHinnant:
@psimon-prnj  Could you test https://github.com/HowardHinnant/date/commits/readlink ?

Thanks.

psimon-prnj:
that works! quick turnaround, nicely done!

HowardHinnant:
Ok, thanks.  That is now pushed to master:  https://github.com/HowardHinnant/date/commits/master

PR #461, opened 2019-05-30
Resolve GCC 9.1.0 noexcept warning

### LINKED ISSUE #459: -Wnoexcept warning with GCC 9.1.0
A -Wnoexcept warning is triggered when compiling with GCC 9.1.0:

```cpp
date-master/src/tz.cpp: In function 'std::istream& date::detail::operator>>(std::istream&, date::detail::MonthDayTime&)':
date-master/src/tz.cpp:965:22: warning: noexcept-expression evaluates to 'false' because of a call to 'constexpr std::chrono::duration<_Rep, _Period>::duration(const _Rep2&) [with _Rep2 = int; <template-parameter-2-2> = void; _Rep = long int; _Period = std::ratio<3600>]' [-Wnoexcept]
  965 |     x = MonthDayTime{};
      |                      ^
```
\
For now I have suppressed it by adding an appropriate noexcept-specification to the default constructor on line 104 of tz_private.h:
```cpp
MonthDayTime() noexcept(false) = default;
```

invexed:
Fixes #459.

### DIFF
--- a/include/date/tz_private.h
+++ b/include/date/tz_private.h
@@ -95,9 +95,9 @@ private:
         U& operator=(const pair& x);
     } u;
 
-    std::chrono::hours           h_{0};
-    std::chrono::minutes         m_{0};
-    std::chrono::seconds         s_{0};
+    std::chrono::hours           h_{};
+    std::chrono::minutes         m_{};
+    std::chrono::seconds         s_{};
     tz                           zone_{tz::local};
 
 public:


HowardHinnant:
Merged: https://github.com/HowardHinnant/date/commit/1f5c192f4a2be7af78d224f37aaaa2726714bb54

Thanks!

PR #464, opened 2019-06-19
Honor symbol visibility settings

rigtorp:
When I use date inside a dynamic library I don't want to export date functions as part of my API so I set default visibility to hidden. This change makes date build honor this setting.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,7 @@
 cmake_minimum_required( VERSION 3.1.0 )
 
+cmake_policy( VERSION 3.12 )
+
 project( date VERSION 2.4.1 )
 
 include( GNUInstallDirs )


HowardHinnant:

rigtorp:
@miketsukerman Given the discussion in #486 can we reopen this PR for review?

HowardHinnant:
Can you rebase it?  Thanks.

HowardHinnant:
Assigned @miketsukerman

miketsukerman:
@HowardHinnant will have a look

rigtorp:
@miketsukerman Rebased

miketsukerman:
@HowardHinnant @rigtorp I would ask you to wait if it's possible with that patch. Cause I want to add ci build for this repository. Otherwise, I'm can't be confident that new changes are not breaking anything.

rigtorp:
@miketsukerman Sure. Do you need help with that? I created the C++ build matrix here which is quiet extensive https://github.com/real-logic/simple-binary-encoding/blob/master/.travis.yml.

miketsukerman:
@rigtorp sure, I'm open for any help. Please have a look at my wip pr https://github.com/HowardHinnant/date/pull/489

HowardHinnant:
@miketsukerman ping.

HowardHinnant:
@qis Do you have an opinion on this PR?  Thanks.

qis:
I have little experience with non-current CMake (usually on nightly), but it looks like this is a reference to https://cmake.org/cmake/help/v3.7/policy/CMP0063.html. Shouldn't it be `3.7` or `CMP0063`, then?

I'm surprised that

> target_link_libraries(my_dynamic_lib **PRIVATE** date::date date::tz)

is not honored unless this is set. If that's the case, then this change is a very good idea.

Just to be sure, this is an ELF issue, where your `.so` file also exports date/tz symbols even though you link "privately" to `tz.a` and use `date.h` functions, correct?

rigtorp:
Yes, I set:
```
set(CMAKE_CXX_VISIBILITY_PRESET hidden)
set(CMAKE_C_VISIBILITY_PRESET hidden)
set(CMAKE_VISIBILITY_INLINES_HIDDEN 1)
```
But these settings are ignored (with a CMake warning) for subprojects in older versions of CMake. Since date specify CMake 3.1 as required, CMake will use the old behavior even on newer versions of cmake.

I specify policy version 3.12 because that also gives support for LTO.

qis:
Isn't LTO something that should be done in the toolchain file? Anyway, great fix!

rigtorp:
@qis https://cmake.org/cmake/help/latest/variable/CMAKE_INTERPROCEDURAL_OPTIMIZATION.html#variable:CMAKE_INTERPROCEDURAL_OPTIMIZATION

Same story as with visibility, not honored for old CMake

PR #466, opened 2019-07-10
Fix iOS build: Include ios.mm when compiling to make iOSUtils available

RunningWithScissors-:
When compiling for iOS the ios.mm file is required in order to have implementaions for functions in iOSUtils (otherwise it doesn't link).

Edit: for some reason it identified me as MarsMero ¯\\_(ツ)_/¯


HowardHinnant:

PR #470, opened 2019-07-22
HAS_UNCAUGHT_EXCEPTIONS now includes C++17

### LINKED ISSUE #435: HAS_UNCAUGHT_EXCEPTIONS not set for C++17 compilers
The test for C++17 was changed in the following commit from `>=` to `>`

https://github.com/HowardHinnant/date/commit/5ba1c1ad8514362dba596f228eb20eb13f63d948#diff-60920d5d6696dfe1f8a446d6e86869b2R130

`std::uncaught_exception()` is deprecated in C++17 and so fails to build cleanly on C++17 compilers such as GCC 8.2.

This version of the code is now live on vcpkg an causes builds using warnings as errors to fail.

Please change this test back to `__cplusplus >= 201703` so that `date.h` builds cleanly on compliant compilers.

rcdailey:
The condition was `__cplusplus >= 201703` which excludes C++17.
However, `std::uncaught_exceptions()` was only introduced in C++17.

Fixes #435

### DIFF
--- a/include/date/date.h
+++ b/include/date/date.h
@@ -137,7 +137,7 @@ namespace date
 #endif
 
 #ifndef HAS_UNCAUGHT_EXCEPTIONS
-#  if __cplusplus > 201703
+#  if __cplusplus >= 201703
 #    define HAS_UNCAUGHT_EXCEPTIONS 1
 #  else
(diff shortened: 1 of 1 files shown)


rcdailey:
Closing due to the author's refusal to accept it upstream.

PR #471, opened 2019-07-26
Fixed deprecation warning on C++17

nike4613:
The `HAS_UNCAUGHT_EXCEPTION` macro definition in `date.h` previously used `> 201703` instead of `>= 201703`, which led to deprecation warnings when building for C++ 17.

### DIFF
--- a/include/date/date.h
+++ b/include/date/date.h
@@ -137,7 +137,7 @@ namespace date
 #endif
 
 #ifndef HAS_UNCAUGHT_EXCEPTIONS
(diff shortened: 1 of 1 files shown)


HowardHinnant:
Thank you for the PR.  However I am respectfully declining this change.

I am intentionally using the deprecated `uncaught_exception` function in C++17.  However if this causes you problems, you can compile with `-DHAS_UNCAUGHT_EXCEPTIONS=1` on the command line.

PR #475, opened 2019-08-06
Test and fixes for zoned_time deduction guides

tomaszkam:
This implements a bunch of test for the zoned_time deduction guides.
This test showed up that a few fixes are necessary:
1) `string_view` constructor still produces better candidates, so the resolution is not reduced
2) we mishandle types that are convertible to `string_view` only as rvalue (suggested by Tim Song)

### DIFF
--- a/include/date/tz.h
+++ b/include/date/tz.h
@@ -709,6 +709,18 @@ using zoned_seconds = zoned_time<std::chrono::seconds>;
 
 #if HAS_DEDUCTION_GUIDES
 
+namespace detail
+{
+   template<typename TimeZonePtrOrName>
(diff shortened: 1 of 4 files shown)


tomaszkam:
Did you have time to check the pull request?

HowardHinnant:
I got half way through my check day before yesterday, and then yesterday was a disaster.  I hope to get back to it today.  Sorry for the delay.

tomaszkam:
Did you find any issues with the code? I would like to land an implementation before submitting the issue.

HowardHinnant:
This is really fine work.  Sorry it took so long to review it.

PR #486, opened 2019-09-06
Improve main CMake file

miketsukerman:
Rename target date_interface to interface and tz to timezone.
Export targets date::interface and date::timezone from cmake.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,7 @@
 cmake_minimum_required( VERSION 3.1.0 )
 
 project( date_prj )
+set(DATE_LIBRARY_VERSION 1.0.0)
 
 include( GNUInstallDirs )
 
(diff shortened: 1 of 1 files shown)


HowardHinnant:
Thanks for doing this.  But I'm going to be removing CMake support in the future due to my inability to maintain it.  I'm looking for volunteers to host CMake support on their own github site which I would gladly point to from my readme.

warzon:
> Thanks for doing this. But I'm going to be removing CMake support in the future due to my inability to maintain it. I'm looking for volunteers to host CMake support on their own github site which I would gladly point to from my readme.

Can I ask what is the reason for removing cmake support?

HowardHinnant:
I don't use it, nor understand it.  So I am unable to help maintain it, or even review quality PR's to improve it such as yours.

miketsukerman:
Making a separate fork with cmake looks not very efficient. May I consider helping you in such cases? So you can leave cmake's in the original repository.

HowardHinnant:
Ok, we'll try it!  You're in charge of reviewing all PR's involving CMakeLists.txt.   I've got to warn you, the pay is lousy. ;-)

PR #488, opened 2019-09-09
Renamed targets and fixed shared library support.

qis:
As discussed in https://github.com/HowardHinnant/date/commit/3a343adf6a40ae290503e23895acb7c132898515, I've renamed the targets to `date` and `tz` to be more consistent with common practices and current Vcpkg target names.

I've also fixed shared library support by setting `DATE_BUILD_DLL` and exporting `DATE_USE_DLL`.

[Here](https://github.com/qis/date-port) is a Vcpkg port replacement to test the changes (tested only on Windows and Linux).

Should I also add an option to disable remote API while `USE_SYSTEM_TZ_DB` is `0` so that the behavior and dependencies are consistent on all platforms when the user wants to download the timezone database files manually?

We currently use `date.h` and `tz.h` in production and ship the timezone database within the `deb` and `msi` files.


qis:
As a side note: The `CMAKE_CXX_STANDARD` setting in `CMakeLists.txt` does not overwrite the higher version set in our toolchain file, so all is good on that front.

miketsukerman:
@qis  I'm completely okay with your changes. But don't have the rights to approve, unfortunately.

HowardHinnant:
Since we're just getting started on our new process, here's what I'd like to do:

I'll assign @miketsukerman everything CMake-related, and I've added a "Passed" flag that can be chosen.  When Mike wants me to merge it, set the Passed flag.  For today I've set that flag myself.

PR #489, opened 2019-09-10
Enable CI build 

miketsukerman:
The first approximation of build matrix for date library. For now, It includes only Linux targets.  
On those targets, build goes fine:
```
Ubuntu 16.04 with gcc 7
Ubuntu 16.04 with gcc 8
Ubuntu 16.04 with gcc 9   

Ubuntu 18.04 with gcc 7

Ubuntu 18.05 with clang 6
```
I've checked on a couple of others. For example, with gcc 4 to 6, there are issues. Also with macOS targets. 

The question is what minimal amount of configurations we need for testing? I mean what operating systems and compiler versions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,66 @@
+language: cpp
+
+matrix:
+  include:
+   
+    - os: linux
+      addons:
+        apt:
+          sources:
+            - ubuntu-toolchain-r-test
+          packages:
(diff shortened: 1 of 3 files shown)


HowardHinnant:
My experience to date with travis has been quite poor:  frequent false failures.  But we'll give this a try again.

PR #500, opened 2019-10-21
Add possibility to remove compilation of current_zone for embedded systems

Boldie:
On embedded systems there is often no filesystem. These things are stored somewhere else and it is easier to feed this directly or create the zone by itself rather than implementing a filesystem. With this little help one can compile the stuff without implementing the current_zone() function. The function can be easily replaced by something specific for the system. If there are better suggestions, I am open for discussion.

### DIFF
--- a/include/date/tz.h
+++ b/include/date/tz.h
@@ -288,7 +288,9 @@ DATE_API const time_zone* locate_zone(std::string_view tz_name);
 DATE_API const time_zone* locate_zone(const std::string& tz_name);
 #endif
 
+#ifndef TZDB_NO_CURRENT_ZONE
 DATE_API const time_zone* current_zone();
+#endif
 
 template <class T>
 struct zoned_traits
@@ -1200,7 +1202,10 @@ struct tzdb
 #else
     const time_zone* locate_zone(const std::string& tz_name) const;
 #endif
+
+#ifndef TZDB_NO_CURRENT_ZONE
     const time_zone* current_zone() const;
+#endif
 };
 
 using TZ_DB = tzdb;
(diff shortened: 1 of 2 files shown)


HowardHinnant:
If there is no filesystem, is there anything at all in tz.cpp that is of service?  I can well imagine that on such a platform tz.h might be useful with custom time zones (e.g. https://github.com/HowardHinnant/date/blob/master/include/date/ptz.h), but there is no need to compile or link to tz.cpp.

Boldie:
After reading a lot of the code in tz.cpp I have the following conclusion:
It is hard to read because there are many ifdefs and maybe it would be favourable to split off the common stuff (database and the evaluation stuff) from the gathering parts. This would make the code much cleaner, easier to understand and as a plus it would be easy for e.g. to feed the code from a string or a flash area. I think this would be not so much work to do, because instead of opening files and iterating over multiple files, only the parsing section is needed to feed the database in init_tzdb. So it would be possible to cat all zonefiles (removing comments first) and feed them as a whole.
-> So this request here can be closed, there is more stuff todo to make it independent of the reading stuff.

HowardHinnant:
Thanks for your comments.

Note that this library should be delivered by your C++ vendor in C++20.  Although I don't know which C++ vendor you use and can't guarantee that this will happen.

Also, here is an example program that uses POSIX time zones and does not need to link against tz.cpp at all:
```
#include "date/ptz.h"
#include <iostream>

int
main()
{
    using namespace date;
    using namespace std::chrono;
    
    zoned_time pzt{Posix::time_zone{"EST5EDT,M3.2.0,M11.1.0"}, system_clock::now()};
    std::cout << pzt << '\n';
}
```
This just output for me:
```
2019-10-21 17:02:49.101876 EDT
```

PR #508, opened 2019-11-16
Revert previous change to cmakelist.txt

### LINKED ISSUE #505: Usage of cmake_policy in CMakeLists.txt is incorrect
At the very least, the way it is used makes the `cmake_minimum_required` call mostly irrelevant, as `cmake_policy` immediately forces the CMake version to be a minimum of 3.14 or greater. I'm not entirely sure what the minimum version of the project should be because of this, but this currently makes the library unusable with a base installation of CMake on Ubuntu 18.04.

beached:
This reverts the changes to CMakeList 1 revisions to allow it to build on systems it used to fix #505

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,52 +1,34 @@
-#[===================================================================[
-   date library by Howard Hinnant
-
-   CMake projects that wish to use this library should consider
-   something like the following :
-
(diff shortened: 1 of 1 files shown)


mellery451:
reverting removes needed functionality.

PR #510, opened 2019-11-20
Work around for NVCC compiler bug

lakshayg:
The NVCC compiler fails to compile the date library due
to a compiler bug which causes it to emit an error when
directly using std::ratio_{multiply,divide} in the template
instantiations for std::duration. This PR works around the
issue by introducing custom ratio_{multiply,divide} which
delegate to the standard library templates

### DIFF
--- a/include/date/date.h
+++ b/include/date/date.h
@@ -158,6 +158,17 @@ namespace date
 #  undef sun
 #endif
 
+// Work around for a NVCC compiler bug which causes it to fail
+// to compile std::ratio_{multiply,divide} when used directly
+// in the std::chrono::duration template instantiations below
+namespace detail {
+template <typename R1, typename R2>
+using ratio_multiply = decltype(std::ratio_multiply<R1, R2>{});
+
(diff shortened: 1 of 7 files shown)


lakshayg:
Hi @HowardHinnant I was able to come up with a slightly better workaround than manually computing the duration periods (https://stackoverflow.com/a/55579846/10923940) and here's a PR for the same. I think it fits much better because the library already has workarounds for GCC and MSVC issues and having this workaround for NVCC will allow a wider audience to use the library without diving into the internals.

I feel this is important because the parts that causes build issues are in the header and it is easy for them to end up in cuda files as a transitive dependency even when the gpu code isn't using the library (as was in my case).

I have also submitted a bug report to Nvidia and hopefully they will fix the issue on their side in some future release.

And finally, thanks for maintaining this library, it has been super useful for me.

HowardHinnant:
Merged: https://github.com/HowardHinnant/date/commit/018a50bcd08bd6a6c5347444785392990fd6bc40

PR #512, opened 2019-11-26
Allow remote_download caller to obtain error message

Ten10:
A bit frustrating when remote_download fails and you need to debug it in assembly just to get an error code which doesn't always help.

This is the most basic change in order to allow the user to get an error message that will help him debug the issue.


HowardHinnant:
Merged: https://github.com/HowardHinnant/date/commit/940f4a5ceb9f45143b32c68f93321fb4d2262c6c

PR #515, opened 2019-11-27
Constrain make_zoned for icc16

nehaljwani:
Build fails with Intel Compiler v16:
```
<source>(1698): error: expression must have a constant value
  std::is_class<typename std::decay<decltype(*std::declval<TimeZonePtr&>())>::type>{}
                                                                                    ^
```


nehaljwani:
I don't have enough C++ know how, but this patch let the build pass.

HowardHinnant:
Merged: https://github.com/HowardHinnant/date/commit/4c951652988b51c6adcf8e95646792a750994829

Issue #504, opened 2019-11-12
PUBLIC_HEADER issue


MaximNM:
Hello,

I'm facing with a PUBLIC_HEADER issue.

CMake Error at CMakeLists.txt:75 (set_target_properties):
  INTERFACE_LIBRARY targets may only have whitelisted properties.  The
  property "PUBLIC_HEADER" is not allowed.

How to resolve it?

HowardHinnant:
Personally I don't use CMake for this library.  It introduces far more problems than it solves.  However there are others that might have some good ideas on how to fix this.  Hopefully they will chime in.

elohim-meth:
Property PUBLIC_HEADERS for INTERFACE_LIBRARY is supported in cmake from version 3.15.
Until that you have to manually install header files:

```cmake
if(${CMAKE_VERSION}` VERSION_GREATER_EQUAL "3.15.0")
    set_target_properties( date PROPERTIES PUBLIC_HEADER include/date/date.h )
endif()

if(${CMAKE_VERSION} VERSION_LESS "3.15.0")
    install(
        FILES "${CMAKE_CURRENT_SOURCE_DIR}/include/date/date.h"
        DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/date )
endif()
```

MaximNM:
@elohim-meth Thanks a lot!

@HowardHinnant Can somebody prepare a patch for it?

HowardHinnant:
Yes please. ;-)

PR #522, opened 2019-12-18
Fix cmake install for 3.14 or earlier (3.7 min)

mellery451:
Fixes: #504, #505


movitto:
Confirmed this fixes the 'date' build (and hence the 'rippled' build) on Fedora 30 (cmake 3.14.5)

Issue #458, opened 2019-05-30
[CMake Regression] USE_SYSTEM_TZ_DB prevents other use cases


emptyVoid:
It's no longer possible to feed ```date``` with custom TZ database while having remote API disabled.

For version ```2.4.1``` I used ```-DCMAKE_CXX_FLAGS="-DAUTO_DOWNLOAD=0 -DHAS_REMOTE_API=0"``` build argument to disable the downloading logic, and then passed custom TZ to ```date::set_install```.

Now this is no longer supported since the following block limits ```date``` to 2 use cases:
https://github.com/HowardHinnant/date/blob/3933a0122d50c0def18f7bb96b2bf643ca7af1fe/CMakeLists.txt#L50-L64

mellery451:
hmm - I don't see the `set_install` interface in the code...maybe it has been changed or removed. I'm not sure if you are using the cmake build, but if there are one or more build options that we can add to support your use-case, I'm happy to try to add it. Maybe describe the specific set of macros you need and we can make it available as one or more build options.

emptyVoid:
Why, it's still there from what I can tell:
https://github.com/HowardHinnant/date/blob/940f4a5ceb9f45143b32c68f93321fb4d2262c6c/src/tz.cpp#L304-L308
https://github.com/HowardHinnant/date/blob/940f4a5ceb9f45143b32c68f93321fb4d2262c6c/include/date/tz.h#L1310

In short my use case is something like this:
```cmake
target_compile_definitions( tz PRIVATE -DUSE_AUTOLOAD=0 )
target_compile_definitions( tz PRIVATE -DHAS_REMOTE_API=0 )
target_compile_definitions( tz PUBLIC -DUSE_OS_TZDB=0 )
```

PR #527, opened 2020-01-04
Add MANUAL_TZ_DB configure option:

mellery451:
FIXES: #458

@emptyVoid would this work for your use-case?


emptyVoid:
Thanks!

Though it doesn't quite work, I'm getting an error:
```
CMake Error at CMakeLists.txt:86 (add_library):
  Target "tz" links to target "CURL::libcurl" but the target was not found.
  Perhaps a find_package() call is missing for an IMPORTED target, or an
  ALIAS target is missing?
```
Changing the linkage check:
https://github.com/HowardHinnant/date/blob/b9cd9c3fde5a8d43c53b60108a998dd945beceba/CMakeLists.txt#L122-L125
to:
```cmake
if( NOT USE_SYSTEM_TZ_DB AND NOT MANUAL_TZ_DB )
```
fixes the issue.

And you might want to add printing of the new option here:
https://github.com/HowardHinnant/date/blob/b9cd9c3fde5a8d43c53b60108a998dd945beceba/CMakeLists.txt#L50-L54

mellery451:
@emptyVoid sorry for the delay - your suggestions implemented in 8878961. Thanks.

emptyVoid:
@mellery451, thank you, works well!

PR #529, opened 2020-01-10
Update Date.h

NexWeb:
Arithmetic overflow: Using operator '-' on a 4 byte value and then casting the result to a 8 byte value. Casting the value to the wider type before calling operator '-' to avoid overflow.

### DIFF
--- a/include/date/date.h
+++ b/include/date/date.h
@@ -1564,7 +1564,7 @@ inline
 month
 operator+(const month& x, const months& y) NOEXCEPT
 {
-    auto const mu = static_cast<long long>(static_cast<unsigned>(x)) + (y.count() - 1);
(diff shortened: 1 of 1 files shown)


include/date/date.h
```diff
@@ -1564,7 +1564,7 @@ inline
 month
 operator+(const month& x, const months& y) NOEXCEPT
 {
-    auto const mu = static_cast<long long>(static_cast<unsigned>(x)) + (y.count() - 1);
+    auto const mu = static_cast<long long>(static_cast<unsigned>(x)) + ((int64_t)y.count() - 1);
```
HowardHinnant: Good find.

How about just removing the parentheses around `y.count() - 1`?  Then the compiler will implicitly do the cast.  I'm not sure why I put those parentheses in there in the first place.

NexWeb: Thank you. Sounds good about removing the parenthesis.

PR #531, opened 2020-01-11
Update date.h

NexWeb:
Arithmetic overflow: Using operator '+' on a 4 byte value and then casting the result to a 8 byte value. Casting the value to the wider type before calling operator '+' to avoid overflow.

### DIFF
--- a/include/date/date.h
+++ b/include/date/date.h
@@ -1718,7 +1718,7 @@ operator<<(std::basic_ostream<CharT, Traits>& os, const year& y)
     detail::save_ostream<CharT, Traits> _(os);
     os.fill('0');
     os.flags(std::ios::dec | std::ios::internal);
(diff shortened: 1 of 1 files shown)


include/date/date.h
```diff
@@ -1718,7 +1718,7 @@ operator<<(std::basic_ostream<CharT, Traits>& os, const year& y)
     detail::save_ostream<CharT, Traits> _(os);
     os.fill('0');
     os.flags(std::ios::dec | std::ios::internal);
-    os.width(4 + (y < year{0}));
+    os.width(4 + int64_t(y < year{0}));
```
HowardHinnant: There's no overflow possible here.  The only two possible expressions are `4 + 0` or `4 + 1`.

NexWeb:
Thank you.

Issue #453, opened 2019-05-21
Benchmarks for timezoned datetimes ?


jkhoogland:
Hi,
I was wondering if there are any benchmarks available for generation of large number of timezoned datetimes ? 
Thanks
Jiri

HowardHinnant:
No, there are not.

Such benchmarks will be very sensitive to a large number of factors, including:

* The precise functionality tested.
* Build settings such as `USE_OS_TZDB` and `AUTO_DOWNLOAD`.

jkhoogland:
fair enough. 
The question came up related to using the timezone information.
I was expecting that if I do something like
```
auto tpUTC = sys_time<nanoseconds>( 123456789 ); // some time point
auto const* tz = date::locate_zone( "America/New_York"); // locate the timezone once

for(int i=0; i!=100000; ++i)
{
   auto const zt = make_zoned( tz, tpUTC );
   auto const tp = zt.get_local_time();
   auto const dp = floor<days>( tp );
   auto const ymd = year_month_day( dp );
}
```
should be faster than
```
auto tpUTC = sys_time<nanoseconds>( 123456789 );

for(int i=0; i!=100000; ++i)
{
   auto const zt = make_zoned( "America/New_York", tpUTC ); // locate the timezone every time
   auto const tp = zt.get_local_time();
   auto const dp = floor<days>( tp );
   auto const ymd = year_month_day( dp );
}
```
But when I run both pieces of code the time used is about the same:
```
time once = 333 millis
time all = 350 millis
```
333 millis for 100,000 equates to about 3 micros per date, which seems a bit much,
even on an 2015 MBP...

HowardHinnant:
If you are not on Windows, can you try `-DUSE_OS_TZDB=1`?

jkhoogland:
Here is some more info (MBP late 2013)

With *USE_OS_TZDB=0*:
```
time once = 758
time all = 667
```
With *USE_OS_TZDB=1*:
```
time once = 12
time all = 32
```
So there is a massive impact when *not* using the system tzdb..

Now some slowdown is expected I guess, 
but the difference seems very large.

Is there any clear explanation for this big difference ?

Erroneous1:
Make sure you're compiling with at least `-O2`

jkhoogland:
this is with `-O3`...

PR #532, opened 2020-01-13
Optimize rule lookup

juan-punchman:
Addresses #453 

# Optimize rule lookup
This somewhat alleviates the speed issues with the `find_rule` function when `USE_OS_TZDB` is not on. In the while loop in `find_rule`, there were some very expensive function calls that make the lookup very slow. I have optimized some of the logic to minimize the amount of calls to slow functions.

### Database IO
Every call of the `find_next_rule` function read the entire database from disk into memory. That function would be called n times in the `find_rule` function. Obviously the database should only be read once from disk per call to the `find_rule` function. I have added an argument in `find_next_rule` to take a reference to the rules database.

### to_sys calculations
At the top of the while loop in the `find_rule` function, there was a call to convert the mdt argument to epoch time. I found that this call is actually pretty computationally expensive. The only value that can change the system time for `mdt` is the `x.save`. What I have done is calculate a base value for tx once, then get the value of tx from that base once per loop. This cuts down a great deal of computation there.

# Benchmarks
Some simple benchmarks show that these changes that I've made improve the runtime for calculating the local time of a zone time. The test I'm preforming use google's gtest library. The test code converts a zoned time to local time n times, then displays the result times. 

For reference, these benchmarks were run on:
```
ROG Zephryus S GX701GWR
OS: arch linux (linux 5.4.10)
CPU: i7-9750H (12) @ 4.500GHz
GPU: NVIDIA GeForce RTX 2070 Mobile
RAM: 32 GB @ 2666 Ghz
```

### Before Optimization
```
[----------] 1 test from date
[ RUN      ] date.bench1
lookup: once      : time used 1048982 us
lookup: every call: time used 619290 us
[       OK ] date.bench1 (1668 ms)
[----------] 1 test from date (1668 ms total)
```

### After Optimization
```
[ RUN      ] date.bench1
lookup: once      : time used 849095 us
lookup: every call: time used 434023 us
[       OK ] date.bench1 (1284 ms)
```

# The benchmark code:

```
TEST( date, bench1 )
{
    using namespace std::chrono;
    using namespace date;

    int n = 10000;

    auto sw = StopWatch();

    // lookup for every call in tz db,,,
    {
        auto tpUTC = std::chrono::system_clock::now();
        auto const* tz = date::locate_zone( "America/New_York" ); // locate the timezone once

        for( int i = 0; i != n; ++i )
        {
            auto const zt  = make_zoned( tz, tpUTC );
            auto const tp  = zt.get_local_time(); // expensive call
            auto const dp  = floor<days>( tp );
            auto const ymd = year_month_day( dp );
        }
    }
    fmt::print( "lookup: once      : time used {} us\n", sw.elapsedUs() );
    sw.restart();

    {
        auto tpUTC = std::chrono::system_clock::now();

        for( int i = 0; i != n; ++i )
        {
            auto const zt  = make_zoned( "America/New_York", tpUTC ); // locate the timezone every time
            auto const tp  = zt.get_local_time();
            auto const dp  = floor<days>( tp );
            auto const ymd = year_month_day( dp );
        }
    }
    fmt::print( "lookup: every call: time used {} us\n", sw.elapsedUs() );
}
```

# TL;DR
This simple optimization sped up the local time conversion by about 23% to 42% (depending on zone lookup).

### DIFF
--- a/src/tz.cpp
+++ b/src/tz.cpp
@@ -1502,10 +1502,9 @@ find_next_rule(const Rule* first_rule, const Rule* last_rule, const Rule* r, dat
 //     r->starting_year() <= y && y <= r->ending_year()
 static
 std::pair<const Rule*, date::year>
-find_next_rule(const Rule* r, date::year y)
+find_next_rule(const Rule* r, date::year y, const std::vector<date::detail::Rule>& rules)
 {
     using namespace date;
-    auto const& rules = get_tzdb().rules;
     if (y == r->ending_year())
     {
         if (r == &rules.back() || r->name() != r[1].name())
@@ -1620,14 +1619,23 @@ find_rule(const std::pair<const Rule*, date::year>& first_rule,
 {
     using namespace std::chrono;
     using namespace date;
+    auto const& rules = get_tzdb().rules;
     auto r = first_rule.first;
     auto ry = first_rule.second;
     sys_info x{sys_days(year::min()/min_day), sys_days(year::max()/max_day),
                seconds{0}, initial_save, initial_abbrev};
+    auto tx_base = mdt.to_sys(y, seconds(0), seconds(0));
+    auto tx_zone = mdt.zone();
     while (r != nullptr)
     {
         auto tr = r->mdt().to_sys(ry, offset, x.save);
-        auto tx = mdt.to_sys(y, offset, x.save);
+
+        auto tx = tx_base;
+        if (tx_zone == tz::standard)
+            tx -= offset;
+        else if (tx_zone == tz::local)
+            tx -= offset + x.save;
+
         // Find last rule where tx >= tr
         if (tx <= tr || (r == last_rule.first && ry == last_rule.second))
         {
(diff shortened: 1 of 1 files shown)


HowardHinnant:
Sorry for the delay in looking at this.

Your comment above about the entire db being read into memory on each iteration is incorrect.  `get_tzdb()` returns a reference to a function local static.  The initialization of this static is what reads the db into memory.  On subsequent calls, `get_tzdb()` simply returns a reference to that same static without re-initializing it.

I'm testing date/test/tz_test/validate.cpp with and without this patch and measuring virtually no difference in the execution times on average.  The variance between runs is much greater than the difference between the average of the runs.  I've tested with `AUTO_DOWNLOAD` set to both 0 and 1 to eliminate network effects.

While testing does appear to confirm your patch is correct (validate.cpp outputs the same answers), I'm hesitant to make this change as I'm not showing a clear benefit.

jmc1283:
Yes, you are right about `get_tzdb()` returning a reference. Nevertheless, only calling it once eliminates some unnecessary function calls.

Maybe I wasn't clear in my initial comment, the function that I have really optimized is `get_local_time`. Since the rule lookup is an n time loop, the speed increase will vary by what time zone you are looking up. Because of that I'm not sure that the validate.cpp are the best tests for this PR, as my profiling of that file shows that only about 1/3 - 1/4 of the work that the validate.cpp is `get_local_time`. Despite this, on my system I saw a pretty significant 200ms difference in validate.cpp's execution time when comparing the master branch to my fork. 

This optimization shows its value when the loop in find_rule has many iterations by reducing unused work for zones with particularly slow lookups (like mine America/New_York).

PR #535, opened 2020-01-16
fixes #534

### LINKED ISSUE #534: CMake fails to link with CURL::libcurl
Building on a clean image of ubuntu:bionic fails with:
```
CMake Error at external/date/CMakeLists.txt:88 (add_library):
  Target "tz" links to target "CURL::libcurl" but the target was not found.
  Perhaps a find_package() call is missing for an IMPORTED target, or an
  ALIAS target is missing?
```

The issue is actually on this line:
https://github.com/HowardHinnant/date/blob/e245f7c6f132c396a9ffc1b81cbed7eb41696dc0/CMakeLists.txt#L124

The CURL::libcurl target wasn't added to FindCURL [until CMake 3.12](https://cmake.org/cmake/help/latest/release/3.12.html#modules).  We could update the minimum CMake version or use `${CURL_LIBRARIES}`, which is still supported in newer versions of CMake.

robertcampion:
Revert from newer target syntax to old library name syntax for compatibility with earlier CMake versions.


mellery451:
LGTM 👍

PR #545, opened 2020-02-22
Fix mingw warnings

HazardyKnusperkeks:


HowardHinnant:
Partially addressed by a2fdba1a.

I made a purposeful decision to use the deprecated `uncaught_exception` in C++17 instead of the new `uncaught_exceptions`.  This decision has satisfied %50.001 of my customers and disappointed the rest.  This choice makes it a little easier for developers to compile with C++17 yet target OS releases prior to C++17 where `uncaught_exceptions` is not supported.  This choice makes it a little harder for developers wishing to compile with warnings about the use of deprecated API's.

My decision can be overridden by compiling with `-DHAS_UNCAUGHT_EXCEPTIONS=1` (or `=0`).  The date library will respect a developer-set value for this macro setting and not attempt to redefine it.

Issue #474, opened 2019-07-30
warning STL4006: std::uncaught_exception() is deprecated in C++17


degski:
y:\vcpkg\installed\x64-windows-static\include\date\date.h(1050,1): warning C4996:  'std::uncaught_exception': warning STL4006: std::uncaught_exception() is deprecated in C++17. It is superseded by std::uncaught_exceptions(), plural.

HowardHinnant:
You can also disable this warning by defining `HAS_UNCAUGHT_EXCEPTIONS=1`.

degski:
Thanks Howard.

degski:
This would be a useful addition to the readme. But I see that the library defines in `date.h:97`:

    #if defined(_MSC_VER) && (!defined(__clang__) || (_MSC_VER < 1910))
    // MSVC
    #  define _SILENCE_CXX17_UNCAUGHT_EXCEPTION_DEPRECATION_WARNING

This means that these warning potentially get suppressed in places I would like to have these warnings. The macro should be undefined at the end of the header, if and when the macro was not defined in the first place. Something like:

    #if !defined(_SILENCE_CXX17_UNCAUGHT_EXCEPTION_DEPRECATION_WARNING)
    #define _SILENCE_CXX17_UNCAUGHT_EXCEPTION_DEPRECATION_WARNING
    #define DATE_UNDEF_SILENCE_CXX17_UNCAUGHT_EXCEPTION_DEPRECATION_WARNING 
    #endif
    
    ....
    
    // at the end.
    #if defined (DATE_UNDEF_SILENCE_CXX17_UNCAUGHT_EXCEPTION_DEPRECATION_WARNING) 
    #undef _SILENCE_CXX17_UNCAUGHT_EXCEPTION_DEPRECATION_WARNING
    #undef DATE_UNDEF_SILENCE_CXX17_UNCAUGHT_EXCEPTION_DEPRECATION_WARNING 
    #endif

grainrigi:
HAS_UNCAUGHT_EXCEPTION is defined as below:
```
#ifndef HAS_UNCAUGHT_EXCEPTIONS
#  if __cplusplus > 201703 // <- excludes C++17
#    define HAS_UNCAUGHT_EXCEPTIONS 1
#  else
#    define HAS_UNCAUGHT_EXCEPTIONS 0
#  endif
#endif  // HAS_UNCAUGHT_EXCEPTIONS
```
[...]

PR #557, opened 2020-03-25
Fix uncaught exceptions availability check

zxey:
I am building this library with `gcc (Ubuntu 9.2.1-9ubuntu2) 9.2.1 20191008` with C++17 and I got lots of compilation warnings, this check seems to be incorrect.

### DIFF
--- a/include/date/date.h
+++ b/include/date/date.h
@@ -138,7 +138,7 @@ namespace date
 #endif
 
 #ifndef HAS_UNCAUGHT_EXCEPTIONS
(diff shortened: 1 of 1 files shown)


HowardHinnant:
Compile with `-DHAS_UNCAUGHT_EXCEPTIONS=1` if you would like to use `std::uncaught_exceptions()` prior to C++20.

zxey:
Sure, I can do that. I thought `std::uncaught_exceptions()` was available since C++17, is there a specific reason why not just enable it?

zxey:
Oh, I see, you answered it here: #474

PR #577, opened 2020-05-07
Use the correct timezone path inside the iOS Simulator

kalliste:
The iOS simulator uses a different path for the timezone database than actual iOS. This patch fixes things for those compiling with -DUSE_OS_TZDB=1 and running under the iOS Simulator, while also not breaking normal iOS.


HowardHinnant:
Merged: d7847666402de4d021faef86199ea7ad81a25c0c

PR #590, opened 2020-07-16
fix ONLY_C_LOCALE export from cmake

### LINKED ISSUE #589: error: operator '!' has no right operand (ONLY_C_LOCALE)
I get some weird errors when trying to compile with date.h:
```
/usr/include/date/date.h:3783:19: error: operator '!' has no right operand
 3783 | #if !ONLY_C_LOCALE
[...]
```
(For full details: https://launchpadlibrarian.net/488835127/buildlog_ubuntu-focal-amd64.waybar_0.9.2-202007161902-0fe0a856-1focal1_BUILDING.txt.gz.)

Seems like the `ONLY_C_LOCALE` is not defined, but I see
```c++
  #ifndef ONLY_C_LOCALE
  #  define ONLY_C_LOCALE 0
  #endif
```
towards the top of `date.h`.

Any idea what might be wrong?

nschloe:
Fixes #589.

The issue with the old code here is that the expansion of the cmake generator expression is that it doesn't resolve to `0` or `1` straight away, as would be needed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -76,10 +76,21 @@ if (CMAKE_VERSION VERSION_GREATER_EQUAL 3.15)
     # public headers will get installed:
     set_target_properties( date PROPERTIES PUBLIC_HEADER include/date/date.h )
 endif ()
+
+if ("${COMPILE_WITH_C_LOCALE}")
+  SET(ONLY_C_LOCALE 1)
+else()
+  SET(ONLY_C_LOCALE 0)
+endif()
+if ("${DISABLE_STRING_VIEW}")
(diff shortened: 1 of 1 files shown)


nschloe:
There are more such things, too:
```
set_target_properties(date::date-tz PROPERTIES
  INTERFACE_COMPILE_DEFINITIONS "USE_OS_TZDB=\$<IF:\$<AND:\$<BOOL:TRUE>,\$<NOT:\$<BOOL:>>,\$<NOT:\$<BOOL:OFF>>>,1,0>;\$<\$<AND:\$<BOOL:>,\$<BOOL:TRUE>>:DATE_USE_DLL=1>"
```

PR #593, opened 2020-08-14
Add Kotlin's datetime library to the list of projects using this

dkhalanskyjb:
The official Kotlin library for handling dates and times is out, and for *nix targets, it uses your library.

As the programmer that was tasked with implementing that library for all supported platforms other than Java and JavaScript, I want to personally thank you for your work: as can be seen here https://github.com/Kotlin/kotlinx-datetime/tree/master/core/nativeMain/cinterop/cpp, the code for *nix (in `cdate.cpp`) is the most succinct, readable and straightforward. Compared to Darwin and WinAPI (with both of which I also had to work as part of this endeavor), you really nailed the interface so that even our non-typical requirements were easily fulfilled!

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,5 +78,6 @@ cmake --build . --target testit # Consider '-- -j4' for multithreading
 * https://github.com/valhalla
 * https://github.com/siodb/siodb
 * https://github.com/KomodoPlatform/atomicDEX-Pro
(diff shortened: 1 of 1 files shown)


HowardHinnant:
Thank you for your kind feedback, and congratulations on Kotlin.

PR #598, opened 2020-08-24
HAS_UNCAUGHT_EXCEPTIONS not defined

neatudarius:
`HAS_UNCAUGHT_EXCEPTIONS` must be defined as 1 for C++17 (201703). See examples at https://github.com/HowardHinnant/date/blob/master/include/date/date.h#L36 and https://github.com/HowardHinnant/date/blob/master/include/date/date.h#L148.

Documentation:
- https://en.cppreference.com/w/cpp/error/uncaught_exception
- https://en.cppreference.com/w/cpp/preprocessor/replace (`__cplusplus | denotes the version of C++ standard that is being used, expands to value 199711L(until C++11), 201103L(C++11), 201402L(C++14), 201703L(C++17), or 202002L(C++20)`

### DIFF
--- a/include/date/date.h
+++ b/include/date/date.h
@@ -138,7 +138,7 @@ namespace date
 #endif
 
 #ifndef HAS_UNCAUGHT_EXCEPTIONS
-#  if __cplusplus > 201703 || (defined(_MSVC_LANG) && _MSVC_LANG > 201703L)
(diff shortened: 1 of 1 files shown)


HowardHinnant:
Please see https://github.com/HowardHinnant/date/search?q=HAS_UNCAUGHT_EXCEPTIONS&type=Issues

Issue #599, opened 2020-08-25
CMake inconsistency build interface pushes date::tz, install interface pushes date::date-tz


matthijs:
Hi,

First the disclaimer ;-) I am by no means an expert in cmake, I use it because it makes my life somewhat easier.

I've installed the date project with the following cmake line:
```bash
cmake -DUSE_SYSTEM_TZ_DB=ON -DBUILD_SHARED_LIBS=ON -DBUILD_TZ_LIB=ON -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_STANDARD=17 -DCMAKE_CXX_FLAGS="-stdlib=libc++" ..
make && sudo make install
```
After that I tried to import the library using
```cmake
find_package(date REQUIRED)
target_link_libraries(main PUBLIC date::tz)
```
This gave an error, after asking on #cmake the target seems to be date::date-tz instead of date::tz. One of the comments is:

the build interface pushes date::tz, the install interface pushes date::date-tz .. one of those is not right

Regards, Matthijs

HowardHinnant:
Personally I don't use (or understand) the cmake build system and use this instead: https://howardhinnant.github.io/date/tz.html#Installation

However perhaps someone here can help you get things going with cmake.

mellery451:
thanks for reporting - I'll provide a fix shortly.

PR #600, opened 2020-08-25
Correct lingering references to bare "tz" libname:

mellery451:
Fixes: #599


mellery451:
@matthijs please let me know if this works for you.

matthijs:
Yes, its working as expected.

PR #601, opened 2020-08-27
fix current_zone detection on buildroot with glibc

paresy:
I supplied the original patch for buildroot with uclibc but now wanted to use glibc, which seems to behave different.
Hopefully this patch should make everyone happy. Let's see what CI will say in minute.

*All the details on why this change was made*

What is wrong: The current_zone is detected as __posix/Europe/Berlin__

When using glibc the proper link is found in /etc/localtime (not on /etc/TZ). This is correct.

I looks like this:
```
# ls -l /etc/localtime
lrwxrwxrwx    1 root     root            39 Aug 27 11:56 /etc/localtime -> ../usr/share/zoneinfo/Europe/Berlin
```

The realpath differs, as the real tzdb resides in the /usr/share/zoneinfo/posix folder:
```
# realpath /etc/localtime
/usr/share/zoneinfo/posix/Europe/Berlin
```

Looking into /usr/share/zoneinfo we get this:
```
# ls -l /usr/share/zoneinfo/
total 48
lrwxrwxrwx    1 root     root            12 Aug 27 08:08 Africa -> posix/Africa
... a lot of more links into posix/* ...
lrwxrwxrwx    1 root     root            10 Aug 27 08:08 Zulu -> posix/Zulu
-rw-r--r--    1 root     root          4463 Feb 20  2019 iso3166.tab
drwxr-xr-x   18 root     root             0 Aug 27 08:08 posix
drwxr-xr-x   18 root     root             0 Aug 27 08:08 right
-rw-r--r--    1 root     root         19397 Mar 10 01:39 zone.tab
-rw-r--r--    1 root     root         17911 Mar 10 01:39 zone1970.tab
```

What is happening:
current_zone() will use the standard glibc detection, which will determine /usr/share/zoneinfo/posix/Europe/Berlin as the realpath. extract_tz_name will then cut after zoneinfo which will leave __posix/Europe/Berlin as the timezone__, which is not correct.

This PR suggests to rework extract_tz_name to reuse the mechanism which is implemented in the TZ detection for uclibc plus add the validation that "zoneinfo" is included in the mentioned path. This new extract_tz_name function could then also be used for the uclibc variant. Then also update the get_tz_dir detection to recognize the buildroot glibc variant.

### DIFF
--- a/src/tz.cpp
+++ b/src/tz.cpp
@@ -3702,8 +3702,9 @@ extract_tz_name(char const* rp)
     if (pos == result.npos)
         throw runtime_error(
             "current_zone() failed to find \"zoneinfo\" in " + string(result));
-    pos = result.find('/', pos);
-    result.remove_prefix(pos + 1);
+    pos = result.find(get_tz_dir());
+    if (pos != result.npos)
+        result.erase(0, get_tz_dir().size() + 1 + pos);
     return result;
 }
 
@@ -3720,8 +3721,9 @@ extract_tz_name(char const* rp)
     if (pos == result.npos)
         throw runtime_error(
             "current_zone() failed to find \"zoneinfo\" in " + result);
-    pos = result.find('/', pos);
-    result.erase(0, pos + 1);
+    pos = result.find(get_tz_dir());
+    if (pos != result.npos)
+        result.erase(0, get_tz_dir().size() + 1 + pos);
     return result;
 }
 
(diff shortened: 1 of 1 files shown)


paresy:
It seems not that easy, as the deducted tz_dir might not be the "realpath" which is send to extract_tz_name. Maybe it's easier to fix this on the buildroot side and make it more "streamlined" to all the other OSes

paresy:
You can use this patch to fix the issue on the buildroot side: https://gist.github.com/paresy/9f6557d444e953b0d2fd9e1bcc761ad2

Issue #161, opened 2017-06-17
Is Android supported?


allidev:
When I tried to build date on Android, it reported a fatal error: 
tz.cpp:128:23: fatal error: wordexp.h: No such file or directory

Does date work on Android?
Thanks!

HowardHinnant:
The intent is to support Android.  But to be honest, this is a one-pony-show and I'm the only pony, and without an Android device to test on.  `<wordexp.h>` is an external utility to expand path shortcuts such as "~", and is not a major dependence on this library.  This is just used to locate the full file path of the downloaded IANA tz database.

I would be very grateful for any workarounds you might be able to supply to locate an alternate default location for the downloaded IANA tz database.  For example on `iOS`, `date::iOSUtils::get_tzdata_path()` is called instead.  I have no idea what this does, as someone else wrote it, but it boils down to providing an alternative `std::string` path to where the IANA database lives.

Once we get past such fundamentals as how the IANA database is located, this library is very portable, and should work on Android (assuming it has `<chrono>` support).

HowardHinnant:
It might be that I should not attempt to `#include <wordexp.h>` if `USE_OS_TZDB == 1`.  This is a theory which has not been tested.  However `USE_OS_TZDB==1` means that the one should use the OS-supplied database,  in which case the library should know where that is, and does not need the aid of `<wordexp.h>`.

I.e. `#include <wordexp.h>` is likely a gratuitous (unnecessary) include when `USE_OS_TZDB == 1`.

allidev:
I am working on a project what can benefit from date library on Android. I will take a crack at this and see what I come up with.

HowardHinnant:
Commit https://github.com/HowardHinnant/date/commit/c5e58fd0156c068b1135fbd275e92d4bb72f63f6 removes the dependence on `<wordexp.h>` when tz.cpp is compile with `-DUSE_OS_TZDB`.

hjmallon:
Bionic is the c library used on andorid. Here is the function which seems to look for the android tzdata.

https://github.com/AOKP/bionic/blob/55495fbdc0ef24212fd108153d784070cd15370d/libc/tzcode/bionic.cpp#L198

HowardHinnant:
Will reopen if this is still a problem _and_ someone is willing to work on it.

molind:
Hi,

Faced the same problem on Android. And solved it with smallest possible fix. I believe it won't break anything. :)

[...]

PR #608, opened 2020-10-08
Proposed fix for issue #161

molind:
Header include should follow the same preprocessor rules as function definition.


HowardHinnant:
Thanks!

PR #609, opened 2020-10-08
Put back handling DISABLE_STRING_VIEW

theirix:
Commit 48f1455 dropped handling DISABLE_STRING_VIEW. This PR put it back as well as setting HAS_DEDUCTION_GUIDES which requires string view as well.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -94,7 +94,7 @@ else()
   target_compile_definitions( date INTERFACE ONLY_C_LOCALE=0 )
 endif()
 if ( DISABLE_STRING_VIEW )
-  target_compile_definitions( date INTERFACE HAS_STRING_VIEW=0 )
+  target_compile_definitions( date INTERFACE HAS_STRING_VIEW=0 -DHAS_DEDUCTION_GUIDES=0 )
 endif()
 
 #[===================================================================[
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -156,6 +156,9 @@ if( BUILD_TZ_LIB )
         target_include_directories( date-tz SYSTEM PRIVATE ${CURL_INCLUDE_DIRS} )
         target_link_libraries( date-tz PRIVATE ${CURL_LIBRARIES} )
     endif( )
+    if( DISABLE_STRING_VIEW )
+      target_compile_definitions( date-tz PRIVATE -DHAS_STRING_VIEW=0 -DHAS_DEDUCTION_GUIDES=0 )
```
mellery451, line 160: HAS_STRING_VIEW is already being set on the main (interface library): https://github.com/HowardHinnant/date/blob/master/CMakeLists.txt#L96-L98. HAS_DEDUCTION_GUIDES is being autodetected based on std version (https://github.com/HowardHinnant/date/blob/master/include/date/tz.h#L91-L97) - is that not sufficient?

theirix: 1. That's right, HAS_STRING_VIEW is properly set on `date` target, it can be ommited. Reworked the patch.

2. There is an edge case when `HAS_STRING_VIEW=0` is set due to CMake `DISABLE_STRING_VIEW` but `HAS_DEDUCTION_GUIDES=1` is set in headers due to autodetection. The compilation will fail for compilers without string view around https://github.com/HowardHinnant/date/blob/master/include/date/tz.h#L714 . For example, it happens with clang 5.0 with libc++ which advertizes itself as C++17 compiler but does not have a proper string_view in libc++. The PR will fix it.

PR #623, opened 2020-11-11
Resolve #618 (Specify TZDB location via CMake)

### LINKED ISSUE #618: CMake issue - add possiility to specify INSTALL location, not only .
it is not possible atm to specify some custom dir for the `INSTALL` define via CMake.
Atm `INSTALL` will always be set to `.` which is not reflecting all use cases obviously. 

Please make it possible to specify the INSTALL location of the iana database via CMake.

connorjak:
I went ahead and renamed `INSTALL` to `TZDB_INSTALL_PATH` to reduce possibility of name conflicts.

NOTE: A common issue may be that you have multiple executables with different relative paths. I resolve that by inserting a copy of the `tzdata` folder at the same relative path from the each executable.

Related: #618

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,6 +35,9 @@ option( ENABLE_DATE_TESTING "Enable unit tests" OFF )
 option( DISABLE_STRING_VIEW "Disable string view" OFF )
 option( COMPILE_WITH_C_LOCALE "define ONLY_C_LOCALE=1" OFF )
 option( BUILD_TZ_LIB "build/install of TZ library" OFF )
+option( USE_CUSTOM_TZDB_INSTALL_PATH "Use a custom path for finding the tzdata folder" OFF )
+# The location of the database is CUSTOM_TZDB_INSTALL_PATH/tzdata (CUSTOM_TZDB_INSTALL_PATH\tzdata on Windows)
+set(CUSTOM_TZDB_INSTALL_PATH ".")
 
 if( ENABLE_DATE_TESTING AND NOT BUILD_TZ_LIB )
     message(WARNING "Testing requested, bug BUILD_TZ_LIB not ON - forcing the latter")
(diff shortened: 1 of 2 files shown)


connorjak:
NOTE: This would require a documentation update for `set_install()` at https://howardhinnant.github.io/date/tz.html

HowardHinnant:
At this time I would rather not change the API, even that of the build macros, unless it is truly urgent.   The rationale for this position is that this is a period of stability for this library while C++ vendors implement and ship its C++20 equivalent.  In the best of all worlds, a year from now this library will be used only for legacy projects where stability is the highest priority.

connorjak:
Got it, I can revert `TZDB_INSTALL_PATH` to `INSTALL`.

HowardHinnant:
Thanks.

connorjak:
I did the revert in the latest commit; I'd say this PR is ready to merge.

CMakeLists.txt
```diff
@@ -96,6 +104,9 @@ endif()
 if ( DISABLE_STRING_VIEW )
   target_compile_definitions( date INTERFACE HAS_STRING_VIEW=0 -DHAS_DEDUCTION_GUIDES=0 )
 endif()
+if ( USE_CUSTOM_TZDB_INSTALL_PATH )
```
derived-coder, line 107: Shouldn't be this in the 
```
if( BUILD_TZ_LIB )
```
block?

I would put it in this line:
```
     if ( USE_SYSTEM_TZ_DB AND NOT WIN32 AND NOT MANUAL_TZ_DB )
         target_compile_definitions( date-tz PRIVATE PUBLIC USE_OS_TZDB=1 )
     else()
          target_compile_definitions( date-tz PUBLIC INSTALL=${USE_CUSTOM_TZDB_INSTALL_PATH} USE_OS_TZDB=0 )
     endif()
```

Se how I removed the other: ` INSTALL=.` together with `USE_OS_TZDB`, that is IMHO an error, will not be considered

connorjak: Oh yeah, I should move it.

PR #628, opened 2020-12-01
Support Android for get current_zone()

ndusart:
Hello.

In Android, the timezone is not stored in **/etc/localtime** or any other file. We have to look for the value of the system property **persist.sys.timezone** instead.

This PR add this check for Android in `current_zone()` function.

### DIFF
--- a/src/tz.cpp
+++ b/src/tz.cpp
@@ -92,6 +92,10 @@
 #  define TARGET_OS_SIMULATOR 0
 #endif
 
+#if defined(ANDROID) || defined(__ANDROID__)
+#include <sys/system_properties.h>
+#endif
+
 #if USE_OS_TZDB
 #  include <dirent.h>
 #endif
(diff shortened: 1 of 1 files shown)


wanfranck:
@ndusart @HowardHinnant is there any plan towards merging it to master branch?

We encountered a `runtime_error` related to `date::current_zone()` on Android device and now thinking around possible workarounds with that.

HowardHinnant:
Does this patch fix your problem?

TheStormN:
@HowardHinnant this patch really fixes the issue for me. Of course it should be noted that the library only works when an external TZ database is provided as it is unable to parse the one provided by Android.

However the PR as it is now have some issues.

1. The `std::string result` is not used, so it can be removed.
2. The ifdef guards should be changed to `#if defined(ANDROID) || defined(__ANDROID__)`

ndusart:
@TheStormN couldn't test it since then, but since this library has been ported in the standard library, isn't this transparently supported by the std::chrono on the NDK ?

TheStormN:
@ndusart The only compilers which have support for time zones are MSVC 2022 and GCC 14. Clang libc++(used in NDK) is still very far from ready, even in their master branch.

ndusart:
Thanks for the information, that's interesting.

I rebased this PR to integrate the suggestions by @TheStormN.

TheStormN:
@HowardHinnant Can we please get this merged, so I can remove at least part of the local patches?

PR #639, opened 2021-01-17
Feature - Setting the binary tzdb directory path at runtime

### LINKED ISSUE #564: Point to a custom compiled tzdb?
I have two questions:

- Is a compiled tzdb cross platform? If I compile it on my Mac, can it be used on Windows? I believe the answer is yes.

- Is there a way to manually specify the location of a compiled tzdb to use? I've looked around but it doesn't seem like it. If there is a way, can it also be used on windows? CCTZ, for example, allows you to specify the location of a compiled tzdb with an environment variable https://github.com/google/cctz/wiki/FAQ#where-is-the-data-used-by-the-cctz-time-zone-library

### LINKED ISSUE #626: Runtime support for setting the binary tzdb directory path
Pre-requisite for #564 

Hi Howard,

I am very motivated to add support for using a binary tzdb on Windows. I hope that that is something that you would also be interested in, and would be willing to add support for, even though you are on a feature freeze.

As a first step, I've added a PR to my own R package (that uses `<date>`) with changes to the tz code to allow for setting the binary tzdb path at runtime. It seems to work pretty well for me with binary files I pulled from CCTZ's github repo. It currently is only allowed for non-Windows platforms, because I have a feeling the binary parser might have to be tweaked for Windows. Nevertheless, this is a required step for Windows support, since I'd have to provide a binary tzdb in a custom location for that platform. It will also be useful for me on non-Windows platforms, as I can standardize on a specific tzdb version across all platforms.

I've documented the changes in the PR, and would love the opportunity to include them as a PR into date if you think they are appropriate. I've tried to be as backwards compatible as possible.
https://github.com/DavisVaughan/civil/pull/6

Let me know what you think!

DavisVaughan:
Closes #626 
A step towards #564 

This PR allows for setting the time zone database directory path at runtime. It does so through a new helper, `set_tz_dir()`, which must be called before the tzdb is initialized if the user has declared that they are going to manually set the path at runtime.

Because the binary tzdb parser does not yet work on Windows, this PR does not enable the runtime setting of the path on Windows. That said, the eventual goal to resolve #564 is to fix the binary tzdb parser on Windows, and then to enable runtime setting of the directory path there. That should bring really nice performance improvements to Windows.

There is a new option, `USE_BINARY_TZDB`. `USE_OS_TZDB` can be thought of as a subset of this. If `USE_BINARY_TZDB` is turned on, but `USE_OS_TZDB` is not, then the client is required to call `set_tz_dir()` before initializing the tzdb, otherwise a runtime error is thrown. If `USE_OS_TZDB` _is_ on, then the automatic detection of the OS tzdb is done instead, and `set_tz_dir()` is not exposed.

If `USE_BINARY_TZDB` is not defined, then it is set to `USE_OS_TZDB` to be compatible with all previous versions of date.

---

There are two additional changes that I felt had to be made to merge this PR.

Previously, `discover_tz_dir()` and `get_tz_dir()` were always defined if on a non-Windows platform, even if `USE_OS_TZDB` was not defined. I considered this a bug. `discover_tz_dir()` is now only defined when `USE_OS_TZDB` is on (otherwise you don't need to discover it), and `get_tz_dir()` is defined if either `USE_OS_TZDB` or `USE_BINARY_TZDB` is defined. It is required for `get_tz_dir()` to be implemented this way to eventually add support for `USE_BINARY_TZDB` on Windows (i.e. we can't just turn off support for it when on Windows, as is currently done).

I have changed the non-Windows implementation of `current_zone()` to only call `get_tz_dir()` when `USE_OS_TZDB` is on. This is a direct result of the previously mentioned change. It seems like `current_zone()` only calls `get_tz_dir()` in embedded systems and is related to a symlink to an OS tzdb path, so it really feels like `USE_OS_TZDB` should have to be active for this to work properly anyways. It would not be sufficient for `USE_BINARY_TZDB` to be on, but `USE_OS_TZDB` to be off, because this could be a custom location that the symlink probably won't point to.

---

To make sure this is working, I turned it on over in the project I am working on. It seems to work nicely for me on my Mac https://github.com/DavisVaughan/clock/pull/49

I'm not very familiar with C++ testing, so feel free to make changes against this PR as you see fit if you'd like to accept it.

### DIFF
--- a/include/date/tz.h
+++ b/include/date/tz.h
@@ -43,8 +43,16 @@
 // required. On Windows, the names are never "Standard" so mapping is always required.
 // Technically any OS may use the mapping process but currently only Windows does use it.
 
+#ifndef USE_OS_TZDB
+#  define USE_OS_TZDB 0
+#endif
+
 #ifndef USE_BINARY_TZDB
-#  define USE_BINARY_TZDB 0
+#  define USE_BINARY_TZDB USE_OS_TZDB
+#endif
+
+#if (!USE_BINARY_TZDB) && (USE_OS_TZDB)
+#  define error "If USE_OS_TZDB is used, then USE_BINARY_TZDB must also be used."
 #endif
 
 #ifndef HAS_REMOTE_API
@@ -82,9 +90,9 @@ static_assert(HAS_REMOTE_API == 0 ? AUTO_DOWNLOAD == 0 : true,
 #  define USE_SHELL_API 1
 #endif
 
-#if USE_BINARY_TZDB
+#if USE_OS_TZDB || USE_BINARY_TZDB
 #  ifdef _WIN32
-#    error "USE_BINARY_TZDB can not be used on Windows"
+#    error "USE_OS_TZDB and USE_BINARY_TZDB can not be used on Windows"
 #  endif
 #endif
 
@@ -1294,12 +1302,14 @@ tzdb_list::cend() const NOEXCEPT
 
 DATE_API tzdb_list& get_tzdb_list();
 
-#if !USE_BINARY_TZDB
-
+#if USE_BINARY_TZDB
+#  if !USE_OS_TZDB
+DATE_API void set_tz_dir(const std::string& tz_dir);
+#  endif // !USE_OS_TZDB
+#else // !USE_BINARY_TZDB
 DATE_API const tzdb& reload_tzdb();
 DATE_API void        set_install(const std::string& install);
-
-#endif  // !USE_BINARY_TZDB
+#endif // USE_BINARY_TZDB
 
 #if HAS_REMOTE_API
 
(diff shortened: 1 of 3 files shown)


pitrou:
I see, so I am right in interpreting that on Windows, one can use `set_install` to point to an existing location of the text form of the IANA database?

> But the binary parser hasn't been extended to support Windows yet

Does it mean that it is known not to work, or just it hasn't been made available?

HowardHinnant:
Hasn't been made available.

Also in the mix:  The latest VS tools implement this library as part of C++20: https://gcc.godbolt.org/z/da3YKeKa8  This includes the IANA database.

DavisVaughan:
> I see, so I am right in interpreting that on Windows, one can use set_install to point to an existing location of the text form of the IANA database?

Right, and that is what I do here, which ends up pointing at a text version of the tzdb that I ship with this R package
https://github.com/r-lib/tzdb/blob/acb04d177fc7344ad09b2c275b950264fea6c194/src/path.cpp#L21

With these compilation flags set:
https://github.com/r-lib/tzdb/blob/main/src/Makevars.win

PR #651, opened 2021-02-26
std::uncaught_exception() is deprecated in C++17, causing build failures

eugenegff:
std::uncaught_exceptions() is available since C++17, and std::uncaught_exception() is already deprecated in C++17, leading to error in "treat warnings as errors" mode in MSVC 2019 in C++17 mode.

### DIFF
--- a/include/date/date.h
+++ b/include/date/date.h
@@ -136,7 +136,7 @@ namespace date
 #endif
 
 #ifndef HAS_UNCAUGHT_EXCEPTIONS
-#  if __cplusplus > 201703 || (defined(_MSVC_LANG) && _MSVC_LANG > 201703L)
+#  if __cplusplus >= 201703 || (defined(_MSVC_LANG) && _MSVC_LANG >= 201703L)
 #    define HAS_UNCAUGHT_EXCEPTIONS 1
 #  else
 #    define HAS_UNCAUGHT_EXCEPTIONS 0


eugenegff:
Note, that attempt to _SILENCE_CXX17_UNCAUGHT_EXCEPTION_DEPRECATION_WARNING is often too late, as it could be used before #include "date/date.h".

eugenegff:
Quite possible that check for MSVC could be done more sophisticated, as MSVC separates version of language and version of the library.

Current check tests *_MSVC_LANG*, but std::uncaught_exception() is deprecated under 

#if _HAS_CXX17 && !defined(_SILENCE_CXX17_UNCAUGHT_EXCEPTION_DEPRECATION_WARNING) \
    && !defined(_SILENCE_ALL_CXX17_DEPRECATION_WARNINGS)

and *_HAS_CXX17* is defined if *_STL_LANG* > 201402L

HowardHinnant:
Use [these instructions](https://docs.microsoft.com/en-us/cpp/build/working-with-project-properties?view=msvc-160#to-create-a-user-defined-macro) to set `HAS_UNCAUGHT_EXCEPTIONS=1`.

eugenegff:
I will fix the problem for myself, but some googling shows that I am not alone. If there are reasons why changing 
#  if __cplusplus > 201703 || (defined(_MSVC_LANG) && _MSVC_LANG > 201703L)
to
#  if __cplusplus >= 201703 || (defined(_MSVC_LANG) && _MSVC_LANG >= 201703L)
is not acceptable - may be adding _HAS_CXX17 check would be OK?
#  if __cplusplus > 201703 || (defined(_MSVC_LANG) && _MSVC_LANG > 201703L) || (defined _HAS_CXX17 && _HAS_CXX17 == 1)
Ability to compile library out-of-the-box without tinkering with a extra preprocessor defines - is a good thing.

HowardHinnant:
I can make exactly half of my clients happy by making this change.  It's the first thing I did years ago.

jwakely:
In theory the `__cpp_lib_uncaught_exceptions` feature test macro should tell you if the newer `std::uncaught_exceptions()` is available. The current MSVC STL does define that.

HowardHinnant:
It isn't the build environment that's the problem.  It is people building on C++17 and shipping on an OS that hasn't been updated to have `std::uncaught_exceptions()` on the OS.

Probably less of a problem today than it was when I first made this decision about 5 years ago.

In any event, the client can still override my decision by compiling with `DHAS_UNCAUGHT_EXCEPTIONS` on the command line.

PR #652, opened 2021-03-01
Add global cleanup to curl object destructor

satishbhat-umd:
Curl resources are leaked without this

### DIFF
--- a/src/tz.cpp
+++ b/src/tz.cpp
@@ -2863,6 +2863,7 @@ struct curl_deleter
     void operator()(CURL* p) const
     {
         ::curl_easy_cleanup(p);
+        ::curl_global_cleanup();
     }
 };
 


nickledeg:
I think the addition of ::curl_global_cleanup() here is causing an error. If I call: date::locate_zone("America/NewYork") and expect the application to download the data from IANA there is the potential for the deleter to be called multiple times, thus causing an exception to be thrown. This is my first day using this brilliant library so I may well be missing something.

HowardHinnant:
I think you're right.  Thanks for pointing this out.  I will get a fix out asap.

nickledeg:
> I think you're right. Thanks for pointing this out. I will get a fix out asap.
Much appreciated Howard.

HowardHinnant:
Fixed: https://github.com/HowardHinnant/date/commit/b8997743033449a996da70c3ac781d2c7937f3b8

Issue #692, opened 2021-08-17
read_long_double conversion warning on ARM 32bit


nico-engels:
Hello! gcc-10 is warning me in this line:

```
../ANPoup/include/date.h:6408:16: error: conversion from ‘long long unsigned int’ to ‘__gnu_cxx::__promote_2<long double, unsigned int, long double, double>::__type’ {aka ‘long double’} may change value [-Werror=conversion]
 6408 |     return i + f/std::pow(10.L, fcount);
      |                ^
../ANPoup/include/date.h:6408:12: error: conversion from ‘long long unsigned int’ to ‘__gnu_cxx::__promote_2<long double, unsigned int, long double, double>::__type’ {aka ‘long double’} may change value [-Werror=conversion]
 6408 |     return i + f/std::pow(10.L, fcount);
      |            ^
cc1plus: all warnings being treated as errors
```

I think this is because `sizeof(long long unsigned int) == sizeof(long double)` in 32-bit arch (https://godbolt.org/z/qqj1Yn1no). It is OK just `static_cast<long double>` the operands? Thanks

HowardHinnant:
Sure, if `static_cast` silences the warning, that would be fine.

PR #693, opened 2021-08-17
static_cast to silent warning implict conversion

nico-engels:
That happen because `sizeof(long long unsigned int) == sizeof(long double)` in 32-bit architectures.


nico-engels:
Closes #692

PR #710, opened 2021-10-19
Allow `set_install()` on Windows to use a file path containing Unicode

### LINKED ISSUE #690: `set_install()` isn't friendly to wide strings on Windows
Hi Howard,

I have a Windows user who is having issues with `set_install()` (full info [here](https://community.rstudio.com/t/cannot-load-clock-on-win11-error-utc-is-not-a-known-time-zone/111477)). The issue is that they unfortunately have an accent mark in their root file path, which I can't have control over. They end up with an error like this:

```r
#> Error: Timezone database not found at "C:/Users/mlyons/OneDrive - COMPANY Namé/Documents-ML/R/win-library/4.1/tzdb/tzdata"
```

I think the issue is that `set_install()` only supports `std::string`, but really `std::wstring` is needed here? For example, I can sort of reproduce this with the following on a Windows computer (compiling with [these defines](https://github.com/r-lib/tzdb/blob/169567f4e35fb8ba8abfa42a16cf24efd53167b7/src/Makevars#L31-L35)):

```cpp
#include <date/tz.h>

void tzdb_test_cpp() {
  std::string path = "C:\\Users\\smwhi\\Documents\\RStudio Projects\\problèm\\tzdata";
  date::set_install(path);
  const date::tzdb& db = date::get_tzdb();
}
```

The `date::get_tzdb()` call will throw the "Timezone database not found" error (notice the accent in the "problem" folder) 

I think I really need to do something like:

```cpp
std::wstring path = L"C:\\Users\\smwhi\\Documents\\RStudio Projects\\problèm\\tzdata"
```

but I can't pass that to `set_install()`.

If you do try and support this, I've also been told that `file_exists()` should use `_waccess()` rather than `_access()` for those wide strings.
https://github.com/HowardHinnant/date/blob/d57d764707050fa85fc3a6f824c666ed4550b1b4/src/tz.cpp#L2839

DavisVaughan:
Closes #690 

This PR ensures that anywhere the install path set by `set_install()` is used, it is carefully converted to UTF-16 and manipulated with wide character helpers when on Windows.

This is a "minimal" PR in the sense that it only adjusts file manipulation of the install path set by `set_install()`. There are other file manipulation functions in date (like the auto download helpers), but this does not adjust those. However, it does provide a few helpers that might be useful in updating those in the future.

This PR is particularly useful if you compile date with:

```
-DINSTALL=dummy \
-DAUTO_DOWNLOAD=0 \
-DHAS_REMOTE_API=0 \
-DUSE_OS_TZDB=0 \
-DONLY_C_LOCALE=1
```

and then use `set_install()` at run time.

---

This PR mainly adds two things:

- `convert_utf8_to_utf16()` is a Windows-only converter from `std::string` to `std::wstring`. We use this right before we utilize any `_w*()` helper.

-  `class file_streambuf` is a new internal class that extends a `std::streambuf`. It is unfortunate that this is required, but sadly `std::ifstream` does not officially support `wstring` file paths. To work around this, we create a custom streambuf that can correctly open these file paths on Windows, and then we utilize that custom streambuf inside `std::istream`. The API for this and `ifstream` are pretty similar, so it was straightforward enough to switch over.

I will admit that I am not 100% sure about how `file_streambuf` is working, and why we only need to overload the `underflow()` method to get it to work. I am following the advice of a few resources that I linked to in the code comments.

---

I tested this locally on a Windows laptop using `set_install()` with a file path containing Unicode and it works well.

The patch is embedded in this PR for my R package, where I have also added a test specifically for this, and it also passes on the Windows CI builds https://github.com/r-lib/tzdb/pull/13. I'd love to have this included upstream!

### DIFF
--- a/src/tz.cpp
+++ b/src/tz.cpp
@@ -198,6 +198,35 @@ namespace
     using co_task_mem_ptr = std::unique_ptr<wchar_t[], task_mem_deleter>;
 }
 
+static
+std::wstring
+convert_utf8_to_utf16(const std::string& s)
+{
+    std::wstring out;
+    const int size = MultiByteToWideChar(CP_UTF8, 0, s.c_str(), -1, NULL, 0);
+
+    if (size == 0)
+    {
+        std::string msg = "Failed to determine required size when converting \"";
+        msg += s;
+        msg += "\" to UTF-16.";
+        throw std::runtime_error(msg);
+    }
+
+    out.resize(size);
+    const int check = MultiByteToWideChar(CP_UTF8, 0, s.c_str(), -1, &out[0], size);
+
+    if (size != check)
+    {
+        std::string msg = "Failed to convert \"";
+        msg += s;
+        msg += "\" to UTF-16.";
+        throw std::runtime_error(msg);
+    }
+
+    return out;
+}
+
 // We might need to know certain locations even if not using the remote API,
 // so keep these routines out of that block for now.
 static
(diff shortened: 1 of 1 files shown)


hsdk123:
Would be great to have tests in the windows CI build for this as well.

HowardHinnant:
Merged with minor tweaks: https://github.com/HowardHinnant/date/commit/529a09267f6f4b1f3ac0ab11864ef452be92dd10

Issue #713, opened 2021-11-01
4 tests failing when using system TZDB with GCC


Tachi107:
When running `testit` from release 3.0.1 with `OPTIONS='-DONLY_C_LOCALE=1 -DUSE_OS_TZDB=1'`, the tests `posix/ptz.pass.cpp`, `solar_hijri_test/parse.pass.cpp`, and `tz_test/zoned_time.pass.cpp` fail with error `terminate called after throwing an instance of 'std::system_error'` - `what():  Unknown error -1`. When testing the latest master, two tests `from tz_test/zoned_time.pass.cpp` fail.

This happens when using GCC 10.3.0 on Debian Testing and GCC 11.2.0 on Debian Unstable, but not with Clang 11.1.0 on Debian Testing.

Tested with 3.0.1 with https://github.com/HowardHinnant/date/commit/052eebaf0086e6bbc5ead01c3f1a8f02496aa701 applied and with latest master.

Tachi107:
The same failures occur when not defining `ONLY_C_LOCALE`, so `USE_OS_TZDB` seems to be the reason why they are failing

HowardHinnant:
Thanks for your report.

On macOS I get also get a failure on posix/ptz.pass.cpp:

Assertion failed: (is_equal(tzi->get_info(tp), tzp.get_info(tp))), function main, file ptz.pass.cpp, line 96.

The binary form of the tzdb has save==0 for "Europe/Dublin" on 2021-01-01, which is incorrect.  The save should be -60min.  This error in the macOS tzdb files is relatively harmless.  It doesn't cause errors in mapping between local time and UTC, but it will cause the C struct tm member tm_isdst to display 0 instead of 1.

I don't know for sure if this is the same error on the platform you are seeing.  And I am not replicating the other two errors you're reporting, and so am not sure what their cause is.  I would not be surprised if they were also due to minor errors in the OS-supplied binary database.  The binary form of the database lacks detailed information on `save`, and only includes whether it is zero or non-zero.

Tachi107:
Do you have any suggestions as how I could debug this?

HowardHinnant:
You can `cd` into a directory and run `testit` from there.  This will run only the tests in and below that directory, shortening test time.

If you can insert print statements in the test to narrow down what line the test is failing on, that can help.  Also, most everything in this library is printable.  For example to gather the information I posted above about ptz.pass.cpp I simply did this:
```c++
[...]

PR #714, opened 2021-11-02
test: use -pthread when $CXX is g++

Tachi107:
Fixes https://github.com/HowardHinnant/date/issues/713


Tachi107:
[`expr`](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/expr.html) should be available in all POSIX systems

PR #744, opened 2022-08-06
Stop using PATH_MAX

Tachi107:
`PATH_MAX` isn't guaranteed to be defined in Posix environments; it is only on systems that have a path length limit, and even in environments where it is defined its usage can lead to issues.

To avoid using `PATH_MAX`, I've made two main changes:

- Where `realpath()` was used, I've changed the code to use its [POSIX.1-2008](https://pubs.opengroup.org/onlinepubs/9699919799/functions/realpath.html)'s new behaviour, where passing a null pointer as the `resolved_name` buffer results in `realpath()` to automatically allocate a buffer large enough to handle the given path, that is returned to the caller. This has been supported for a long time as a GNU libc extension before being standardized.
- Where `readlink()` was used, the size of the buffer was already determined when calling `lstat()`; the returned `struct stat` contains a `st_size` field, containing the number of bytes needed to store the symbolic link contents. This meant that to avoid using the tricky define I only needed to use a `calloc()`'d buffer instead of a static one, of size `st_size + 1` as `st_size` doesn't account for the null terminator of C strings.

To make sure that memory is always freed, I've wrapped the new dynamic allocations in an `std::unique_ptr` with a deleter that calls `free()` on destruction.

To read more about why `PATH_MAX` leads to buggy code I'd suggest reading something like this: [eklitzke.org/path-max-is-tricky](https://eklitzke.org/path-max-is-tricky)

### DIFF
--- a/src/tz.cpp
+++ b/src/tz.cpp
@@ -484,9 +484,10 @@ discover_tz_dir()
     if (!(lstat(timezone, &sb) == 0 && S_ISLNK(sb.st_mode) && sb.st_size > 0))
         throw runtime_error("discover_tz_dir failed\n");
     string result;
-    char rp[PATH_MAX+1] = {};
-    if (readlink(timezone, rp, sizeof(rp)-1) > 0)
-        result = string(rp);
+    unique_ptr<char[]> rp(new char[sb.st_size]);
+    const auto rp_length = readlink(timezone, rp.get(), sb.st_size);
+    if (rp_length > 0)
+        result = string(rp.get(), rp_length); // readlink doesn't null-terminate
     else
         throw system_error(errno, system_category(), "readlink() failed");
     auto i = result.find("zoneinfo");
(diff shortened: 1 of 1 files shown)


Tachi107:
This patch is currently being deployed in the howardhinnant-date Debian package, and allows tests to pass on 20 different architectures, including GNU Hurd which doesn't define `PATH_MAX`: <https://buildd.debian.org/status/package.php?p=howardhinnant-date>

basilgello:
@Tachi107 wow, you packaged the libdate! Wonder how could I miss it!

Tachi107:
@basilgello nice to hear! Please let me know if you find any issues with it :)

basilgello:
@Tachi107 Can you please upload the package to bullseye-bpo?

Tachi107:
I don't have the permission to do it yet, but while trying to check if doable I've discovered a small issue (see https://github.com/HowardHinnant/date/pull/754#issuecomment-1361248007). Once the issue is fixed and I'm allowed to backport packages, the package will be in bullseye-backports :)

basilgello:
LGTM @HowardHinnant

PR #752, opened 2022-09-26
build: introduce `ENABLE_DATE_INSTALL` to allow user opt-out of install

### LINKED ISSUE #751: CMake: Allow user to opt out of install
Hi, thanks for the great work!

It would be very nice to make this part optional: 
https://github.com/HowardHinnant/date/blob/22ceabf205d8d678710a43154da5a06b701c5830/CMakeLists.txt#L161-L202 

For example when installing a shared library, using `date` as a private dependency, we don't want to install `date`.
A new option should be introduced `ENABLE_DATE_INSTALL` to match project convention, or `DATE_ENABLE_INSTALL` to match cmake project conventions.

Would you consider such PR? Make `ENABLE_DATE_INSTALL` default to `ON` shouldn't change any behavior, and it would give users using `date` with `add_subdirectory` to opt-out of install.

Have a nice day.

OlivierLDff:
When using `date` as a PRIVATE dependency, it is not required to install it. This flag give user using `date` with add_subdirectory the opportunity to disable install target Default behavior is conserved since ENABLE_DATE_INSTALL is ON

fix #751

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,7 +33,7 @@ option( ENABLE_DATE_TESTING "Enable unit tests" OFF )
 option( DISABLE_STRING_VIEW "Disable string view" OFF )
 option( COMPILE_WITH_C_LOCALE "define ONLY_C_LOCALE=1" OFF )
 option( BUILD_TZ_LIB "build/install of TZ library" OFF )
-option( ENABLE_DATE_INSTALL "Enable unit tests" ON )
+option( ENABLE_DATE_INSTALL "Enable install" ON )
 
 if( ENABLE_DATE_TESTING AND NOT BUILD_TZ_LIB )
     message(WARNING "Testing requested, bug BUILD_TZ_LIB not ON - forcing the latter")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,8 +20,6 @@ cmake_minimum_required( VERSION 3.7 )
 project( date VERSION 3.0.1 )
 set(ABI_VERSION 3) # used as SOVERSION, increment when ABI changes
 
-include( GNUInstallDirs )
-
 get_directory_property( has_parent PARENT_DIRECTORY )
 
 # Override by setting on CMake command line.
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -35,12 +33,17 @@ option( ENABLE_DATE_TESTING "Enable unit tests" OFF )
...
+option( ENABLE_DATE_INSTALL "Enable unit tests" ON )
 
 if( ENABLE_DATE_TESTING AND NOT BUILD_TZ_LIB )
     message(WARNING "Testing requested, bug BUILD_TZ_LIB not ON - forcing the latter")
     set (BUILD_TZ_LIB ON CACHE BOOL "required for testing" FORCE)
 endif( )
 
+if( ENABLE_DATE_INSTALL )
+  include( GNUInstallDirs )
+endif( )
```
Tachi107, line 45: Wouldn't it make sense to move this `include()` in the installation section?

OlivierLDff: No because we want CMAKE_INSTALL_INCLUDIR to be defined for target_include_directories

Tachi107: Right, but the variable is not currently used in `target_include_directories()`; I just saw #753 though, and they seem to complement each other. Thanks!

OlivierLDff: Yes sorry for not making it clearer, I split bug fix and feature in different pr to make review easier/faster

CMakeLists.txt
```diff
@@ -35,12 +33,17 @@ option( ENABLE_DATE_TESTING "Enable unit tests" OFF )
 option( DISABLE_STRING_VIEW "Disable string view" OFF )
 option( COMPILE_WITH_C_LOCALE "define ONLY_C_LOCALE=1" OFF )
 option( BUILD_TZ_LIB "build/install of TZ library" OFF )
+option( ENABLE_DATE_INSTALL "Enable unit tests" ON )
```
gerhardol, line 36:
```suggestion
option( ENABLE_DATE_INSTALL "Enable install" ON )
```

gerhardol (approved):
Works as intended, just copy&paste to fix.
I was planning to create a patch like this myself.

basilgello:
LGTM @HowardHinnant

PR #754, opened 2022-09-26
build: don't write to global variable `CMAKE_CXX_STANDARD`

OlivierLDff:
Instead use modern `target_compile_features` with `cxx_std_11` (since this project is compatible with C++11 and later). More information: https://cmake.org/cmake/help/latest/command/target_compile_features.html

If user wants to access c++14/17/20 features, it is up to him to define cxx_std_<version> on his target. Or he can still use the old school CMAKE_CXX_STANDARD global variable.

This commit might break project relying on `date` writing to a global variable. If user links his target to date, his target will be promoted at least to c++11 if not specified otherwise.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,7 +15,7 @@
 
 #]===================================================================]
 
-cmake_minimum_required( VERSION 3.7 )
+cmake_minimum_required( VERSION 3.8 )
 
 project( date VERSION 3.0.1 )
 set(ABI_VERSION 3) # used as SOVERSION, increment when ABI changes
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,9 +24,6 @@ include( GNUInstallDirs )
 
 get_directory_property( has_parent PARENT_DIRECTORY )
 
-# Override by setting on CMake command line.
-set( CMAKE_CXX_STANDARD 17 CACHE STRING "The C++ standard whose features are requested." )
-
 option( USE_SYSTEM_TZ_DB "Use the operating system's timezone database" OFF )
 option( MANUAL_TZ_DB "User will set TZ DB manually by invoking set_install in their code" OFF )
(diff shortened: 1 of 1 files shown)


ChrisThrasher (changes requested):
I like this! I recently did the same thing in fmtlib [here](https://github.com/fmtlib/fmt/pull/3205). Don't forget you need to bump the minimum CMake version to 3.8 to ensure that this compile feature exists.

OlivierLDff:
> I like this! I recently did the same thing in fmtlib [here](https://github.com/fmtlib/fmt/pull/3205). Don't forget you need to bump the minimum CMake version to 3.8 to ensure that this compile feature exists.

Sure done. Have a nice day.

ChrisThrasher (approved):

Tachi107:
> If user wants to access c++14/17/20 features, it is up to him to define cxx_std_ on his target.

Unfortunately it is not that simple. If you compile libdate-tz in e.g. C++17 mode you cannot use it in projects targeting an older standard, as some functions are conditionally complied depending on whether std::string_view is available or not, like `locate_zone()`. The same applies the other way around; if you compile the .cpp file in C++11 mode and then use the headers in C++17 mode, you're code will try to link to the implementation that uses std::string_view, but the complied object only contains the std::string one so you'll get linking errors.

It might make sense to set `cxx_std_17` as PUBLIC; that's what I'm planning to do in the Debian package.

basilgello:
Maybe a crazy idea but… Can we compile both sets of functions somehow to produce libraries with string_view and string? Then we could ship the CMake configutations for every combination of parameters. I did this for kissfft, it is time consuming but doable.

Tachi107:
> Can we compile both sets of functions somehow to produce libraries with string_view and string?

Yes, it should be possible. But even then, the C++ version has to be set to cxx_std_17, otherwise the std::string_view variant cannot be possibly compiled into the library.

PR #755, opened 2022-09-26
build: disable logs when `date` is not the main CMake project

OlivierLDff:
When including `date` with `add_subdirectory`, users logs are polluted with config from `date`. Using `PROJECT_IS_TOP_LEVEL` would be even more appropriate (https://cmake.org/cmake/help/latest/variable/PROJECT_IS_TOP_LEVEL.html#variable:PROJECT_IS_TOP_LEVEL) but it require CMake v3.21 as minimal version.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,6 +17,11 @@
 
 cmake_minimum_required( VERSION 3.7 )
 
+set(DATE_MAIN_PROJECT OFF)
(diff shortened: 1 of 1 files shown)


basilgello:
Why not use `ExternalProject_Add` instead?

OlivierLDff:
Because `add_subdirectory` is what is called internally when using `FetchContent` or  [CPM.cmake](https://github.com/cpm-cmake/CPM.cmake)

PR #769, opened 2023-01-07
docs: support dark color-scheme

Tachi107:
This patch updates the html documentation to add the color-scheme meta element, signalling light and dark color scheme support, and updating the inline CSS styles to override some colors when dark mode is active, to make sure that the document is always readable.

I've also cleaned up a bit the CSS styles, but without functional changes.

I've only updated the latest standard revision (r7), but please let me know if I should revert changes to that file.

Here's how it looks on my system; the dark scheme is provided by the browser:

image

image

PS: to be clear, the page is rendered in dark mode only if the user has set dark mode in the browser/system settings

### DIFF
--- a/chrono_io.html
+++ b/chrono_io.html
@@ -4,20 +4,17 @@
 <head>
 	<title>chrono_io</title>
 
+	<meta name="color-scheme" content="light dark" />
 	<style>
-	p {text-align:justify}
-	li {text-align:justify}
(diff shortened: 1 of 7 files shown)


basilgello:
Thanks for the improvement @Tachi107 ! Helps during night debugging sessions :)

@HowardHinnant LGTM

PR #778, opened 2023-03-01
slow init of vector of (tz) dates

reder2000:
Creating a large vector of localized dates is painfully slow, because a lookup is needed each time (especially in debug mode).
Caching the default result solves the issue.

### DIFF
--- a/include/date/tz.h
+++ b/include/date/tz.h
@@ -293,7 +293,8 @@ struct zoned_traits<const time_zone*>
     const time_zone*
     default_zone()
     {
-        return date::locate_zone("Etc/UTC");
+      static const time_zone* result = date::locate_zone("Etc/UTC");
+      return result;
     }
 
 #if HAS_STRING_VIEW


HowardHinnant:
Unfortunately this patch breaks a use case:  If the remote tzdb API is used to push a new database to the front of the list, followed by the erasure of the original database, that could leave the static `result` pointer dangling.

As a workaround, I recommend initializing a `vector<zoned_time<Duration>>` with `v(N, zoned_time<Duration>{})`.  This gives the same effect, only making one call into the database.

reder2000:
I see your point. But imagine the same thing on a multithreaded context, `v(N)` could be filled with possibly many "default" values.
Choosing a default that is not constant (or constexpr) looks like 
- a bad design decision
- something one would not dare explain to somebody genuine to C++
Also, there are contexts in which one does not control how those are done. Like serialization, where I had to override for vector<zoned_time>, otherwise the thing (nearly) comes to a halt.
I understand the DB may change in the context of a long running program. But then would not all the tz *pointers be invalid ?

Thanks anyway, I'll try to track those inits (if I can).

HowardHinnant:
Here is a more in-depth description of those multithread issues: https://github.com/HowardHinnant/date/wiki/Examples-and-Recipes#tzdb_manage

The bottom line is that there is no one-size-fits-all solution, and the solution is left up to the client.  There is only one call that can invalidate pointers into the database:  `erase_after`, and this library never calls it.  The article explores options on when the client might call it, if ever.

And to fully enable this feature, this library needs to be careful about avoiding long-lived pointers or references into the database.

PR #780, opened 2023-03-03
Fix compilation when MISSING_LEAP_SECONDS is set to true

techtim:
When define MISSING_LEAP_SECONDS throws compile time `error: 'load_just_leaps' was not declared in this scope`
when find_read_and_leap_seconds tries to call load_just_leaps which surrounded with #if !MISSING_LEAP_SECONDS 
This commit surrounds call to load_just_leaps

### DIFF
--- a/src/tz.cpp
+++ b/src/tz.cpp
@@ -2820,6 +2820,7 @@ find_read_and_leap_seconds()
         }
         return leap_seconds;
     }
+#if !MISSING_LEAP_SECONDS
     in.clear();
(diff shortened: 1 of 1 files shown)


HowardHinnant:
What are the symptoms that are being fixed?  Compile-time error?  Run-time error?

I would've guessed that if this code couldn't find leap second information it would just return the empty vector.

techtim:
compile time error: 'load_just_leaps' was not declared in this scope

HowardHinnant:
Thanks!

PR #787, opened 2023-05-20
detect current zone on OpenWRT systems

paresy:
Similar to buildroot we might want to detect the timezone on OpenWRT systems which are also quite popular. (I am not sure if i remember correctly that you didn't really want more systems in the detection part - if so, i just want to provide it for anyone that searches it, so he/she/it can apply a nice ready-to-go patch.)

Thanks for the great library!

### DIFF
--- a/src/tz.cpp
+++ b/src/tz.cpp
@@ -3933,6 +3933,25 @@ sniff_realpath(const char* timezone)
 const time_zone*
 tzdb::current_zone() const
 {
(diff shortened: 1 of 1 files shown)


basilgello:
LGTM @HowardHinnant - just thinking if OpenWRT support should not be the very first option given the file is queried on every operation?

HowardHinnant:
Merged: dc9d161607137f00c1747477c7de7406dc97f5fa

PR #804, opened 2023-10-18
Fixed build for WinRT and Android

eugenegff:
1. Code under #ifdef WINRT uses convert_utf8_to_utf16() function too, so this func should not be defined under #ifndef WINRT

2. Android has no curl.h, wordexp.h and even tz database is in non-standard binary format, but library still could be used on Android with user supplied tz database, if some defines are set to more reasonable defaults.

### DIFF
--- a/include/date/tz.h
+++ b/include/date/tz.h
@@ -49,7 +49,7 @@
 
 #ifndef HAS_REMOTE_API
 #  if USE_OS_TZDB == 0
-#    ifdef _WIN32
+#    if defined _WIN32 || defined __ANDROID__
 #      define HAS_REMOTE_API 0
 #    else
 #      define HAS_REMOTE_API 1
--- a/src/tz.cpp
+++ b/src/tz.cpp
@@ -118,6 +118,9 @@
 // the current time zone. On Win32 windows.h provides a means to do it.
 // gcc/mingw supports unistd.h on Win32 but MSVC does not.
 
+#ifdef __ANDROID__
+#  define INSTALL .
+#endif
 #ifdef _WIN32
 #  ifdef WINAPI_FAMILY
 #    include <winapifamily.h>
--- a/src/tz.cpp
+++ b/src/tz.cpp
@@ -183,20 +183,6 @@ static CONSTDATA char folder_delimiter = '/';
 #if !USE_OS_TZDB
 
 #  ifdef _WIN32
-#    ifndef WINRT
-
-namespace
-{
-    struct task_mem_deleter
-    {
-        void operator()(wchar_t buf[])
-        {
-            if (buf != nullptr)
-                CoTaskMemFree(buf);
-        }
-    };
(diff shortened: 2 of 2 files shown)


basilgello:
@eugenegff Do we need date library on WinRT in 2024 when MSVC 2019 VS 17.0 supports `std::chrono::time_zone` woth C++20 natively? @HowardHinnant  recommends using C++20 chrono in all new code. Thus we could drop WinRT at all. Or you have some usecases for date library under WinRT specifically?

> Android has no curl.h, wordexp.h and even tz database is in non-standard binary format

On Android, I have prepared tzdata parser patch allowing the usage of updatable zoneinfo from Android. But apart of it, disabling `HAS_REMOTE_API` makes sense to me. In Kodi, we use Curl anyway so we never encountered  the missing dependency issues, but general audience will unlikely use curl for that purpose.

eugenegff:
I use this date library for project livehome3d.com built for macOS, iOS, Android and UWP Windows, in C++17 mode. Last time I checked it was not yet possible to switch to C++20 mode on all those platforms. Moreover, at least Apple often limits new library feature’s availability only to latest versions of OSes, while I need to support wider range. Learned it hard way with <filesystem> and float <charconv>.

basilgello:
I managed to solve this problem by letting Windows and UWP to use C++20 features while building date if these are not provided by platform: https://github.com/basilgello/xbmc/commit/5a05aa2c4e4ae7f416803ae23efbc725840429e2

I am not saying this solution fits your project (and I see no obstacles in merging the PR!) but maybe it helps you and other people in the migration efforts.

src/tz.cpp
```diff
@@ -118,6 +118,9 @@
 // the current time zone. On Win32 windows.h provides a means to do it.
 // gcc/mingw supports unistd.h on Win32 but MSVC does not.
 
+#ifdef __ANDROID__
+#  define INSTALL .
```
basilgello, line 122: Previoysly I did the same thing in Kodi but speaking of the upcoming version tag - #752 brings CMake option to opt out of installation so this hunk becomes redundant

include/date/tz.h
```diff
@@ -49,7 +49,7 @@
 
 #ifndef HAS_REMOTE_API
 #  if USE_OS_TZDB == 0
-#    ifdef _WIN32
+#    if defined _WIN32 || defined __ANDROID__
```
basilgello, line 52: I also have somewhat mixed feelings about this #ifdef in general (not your addition of Android):

* On one side, Android has no curl built-in, unlike MacOS,
* On the other, `HAS_REMOTE_API` is also a CMake option, and these kind of defaults only mask the explicit configuration intent

@HowardHinnant WDYT?

HowardHinnant: I'm ok with this.

basilgello:
Given no objections from Howard, LGTM from me!

PR #806, opened 2024-01-12
Prepending CMake options dith DATE_

pbosetti:
This way, when the library is integrated into another project e.g. with CMake FetchContent, options for date lib are cleanly grouped together in ccmake or make-gui


basilgello:
I like the overall idea but for a minor bump that might be considered a breaking change. An alternative might be to check if "old" settings are used and issue a CMake Warning stating the deprecation.

@HowardHinnant what is your opinion?

HowardHinnant:
I don't want to make any breaking changes.  This library is obsoleted by C++20 and should slowly fade away with as little noise as possible.

PR #807, opened 2024-01-13
Add ability to specify timezone database location via environment variable

### LINKED ISSUE #788: Run time specification of tzdb location with environment variable
Hi Howard/all,

# Summary

To facilitate package management, I think it would be very useful to add an environment variable specifying the location of either a binary _or_ a source database, which is read at runtime to discover the location of the database. This has the following benefits:
* Allows a user to keep their timezone database up to date without considering re-compilation
* Allows a user to be flexible about where the timezone database is stored
* Allows `date` to be used portably without compilation of paths into the resulting library (when not used as header only)

# Context

I'm currently in the process of uplifting the conan `date` package to use the newly added `tz` package, which packages the tzdb. Currently it only supports a binary package, but can be uplifted to also package the source distribution.

The challenge I'm facing is that conan stores it packages in the conan cache, `~/.conan/p`. Conan provides the facilities to make these packages available at runtime to consumer libraries via environmental scripts, which would be a convenient way to manage the tzdb to ensure that it remains modern, and distribute it with software releases. The challenge is that there doesn't seem to be a flexible way to tell `date` that this is where the tzdb lives.

To illustrate this, I thought I'd enumerate the ways that currently exist to customise tzdb location and talk about why these are unsuitable from a packaging perspective.

## `USE_SYSTEM_TZ_DB=1`
* This will set `AUTO_DOWNLOAD=0` and `HAS_REMOTE_API=0` to ensure that what's available locally is used
* Will also set `INSTALL=.` and `USE_OS_TZDB=1`, which will use the compiled variant of tzdb in the system installation location
* From a package management perspective, this does most of the right things, except that it doesn't appear as though I can define the location of the database. It looks specifically at the following locations, as far as I can tell:
    * `/usr/share/zoneinfo`
    * `/usr/share/zoneinfo/uclibc`
    * `/var/db/timezone/zoneinfo`
    * `/etc/localtime`

### The problem for package management
Because the location of the database can't be defined, it's not possible to use this to point at a portable version of a compiled database in a custom location.

I could look at attempting to install `tzdb` from conan into the system locations here, but I don't feel this would be wise - this is the domain of system package management and I'm not sure what effect this would have on system package managers.

## `MANUAL_TZ_DB=1`
* This will set `AUTO_DOWNLOAD=0` and `HAS_REMOTE_API=0` to ensure that what's available locally is used
* Will also set `USE_OS_TZDB=0` to ensure that the system database is not used, if present
* This expects the user will either:
     * invoke `set_install(const std::string& s)` in their code to set the location of the database. My assumption is that this refers to a _source_ database only, and this can't be configured for a _binary_ database.
     * Pass a value to `INSTALL=` when compiling, which means that `${INSTALL}/tzdata` is searched for the source database
     * Populate `~/Downloads/tzdata` themselves

### The problem for package management
* Relies on the user to set something that can be reasonably captured as a property of the `date` recipe - adds additional boilerplate
* Only allows setting for a source version of the database
* If the user actually specifies a path here, this would create a non-portable binary because it contains a path unlikely to exist in the deployment environment

##  `MANUAL_TZ_DB=0` and `USE_SYSTEM_TZ_DB=0`
* Will set `AUTO_DOWNLOAD=1`, `HAS_REMOTE_API=1`, `USE_OS_TZDB=0`
* Will add a dependency on curl and attempt to download the latest version of the database src to `~/Downloads/tzdata`

### The problem for package management
Offloads management of tzdata as a dependency to the `date` library, rather than managed by the dependency (package) manager. Not a problem in and of itself, but orthogonal to the philosophy of using a package manager to manage dependencies.

# Related issues

Customisation/management of these locations seems like an oft-requested feature:
* https://github.com/HowardHinnant/date/issues/541
* https://github.com/HowardHinnant/date/pull/611
* https://github.com/HowardHinnant/date/issues/305
* https://github.com/HowardHinnant/date/issues/472
* https://github.com/HowardHinnant/date/issues/397
* https://github.com/HowardHinnant/date/issues/626
* https://github.com/HowardHinnant/date/issues/564
* https://github.com/HowardHinnant/date/pull/639

These seem to have been stagnant for a few years now - in particular #611 seems to be the closest to what would be required here. Is there any intention to add this kind of customisation to the `date` library or is this kind of thing not being considered now that there's been acceptance of this into stdlib?

samuel-emrys:
This adds the ability for a user to populate the TZDATA environment variable to specify the location in which to look for the timezone database, which allows for a runtime query of this location. Without this, date currently only supports compile time determination of the database location.

This facilitates the ability to:
* Use a timezone database provided by a package manager
* Keep the timezone database up to date without re-compilation
* Be flexible about where the timezone database is stored

This is a simplification of https://github.com/HowardHinnant/date/pull/611, removing windows support for a zic-compiled binary database, so that it only adds environmental variable features.

Closes #788

### DIFF
--- a/src/tz.cpp
+++ b/src/tz.cpp
@@ -81,6 +81,9 @@
 #  endif  // __MINGW32__
 
 #  include <windows.h>
+#  if !defined(S_ISDIR) && defined(S_IFMT) && defined(_S_IFDIR)
+#    define S_ISDIR(m) (((m) & S_IFMT) == _S_IFDIR)
+#  endif
 #endif  // _WIN32
 
 #include "date/tz_private.h"
@@ -411,6 +414,14 @@ access_install()
     #undef STRINGIZE
 #endif  // !INSTALL
 
+    {
+        static char* tz_local_env = getenv("TZDATA");
+        if (tz_local_env != nullptr) {
+            static std::string tz_local_env_s = tz_local_env;
+            return tz_local_env_s;
+        }
+    }
+
     return install;
 }
 
@@ -464,6 +475,14 @@ discover_tz_dir()
 #  ifndef __APPLE__
     CONSTDATA auto tz_dir_default = "/usr/share/zoneinfo";
     CONSTDATA auto tz_dir_buildroot = "/usr/share/zoneinfo/uclibc";
+
+    {
+        static char* tz_local_env = getenv("TZDATA");
+        if (tz_local_env != nullptr) {
+            static std::string tz_local_env_s = tz_local_env;
+            return tz_local_env_s;
+        }
+    }
 
     // Check special path which is valid for buildroot with uclibc builds
     if(stat(tz_dir_buildroot, &sb) == 0 && S_ISDIR(sb.st_mode))


szaszm:
Thank you, this would make my life easier as a Nix and NixOS user, where the timezone database is at /etc/zoneinfo. More details: https://stackoverflow.com/questions/3896587/are-linuxs-timezone-files-always-in-usr-share-zoneinfo

basilgello:
Leaving this up to @HowardHinnant because NixOS uses pretty new compilers that support C++20 calendars and time zones natively

samuel-emrys:
This ticket had nothing to do with NixOS except someone finding utility in it using NixOS - not sure why that means it's not relevant anymore.

PR #813, opened 2024-02-17
Allow option() override by FetchContent

### LINKED ISSUE #812: Cannot override CMake options when fetching the library via FetchContent_Declare()
CMake 3.13 added support for adjusting option() variables of libraries which are fetched by `FetchContent_Declare()`. However, the `date` library CMakeLits.txt file does not support that and so an ugly hacks are needed to make it work. The fix is really trivial and I will submit a PR about that.

TheStormN:
Fixes #812

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,6 +24,11 @@ include( GNUInstallDirs )
 
 get_directory_property( has_parent PARENT_DIRECTORY )
 
+if (POLICY CMP0077)
+    # Allow CMake 3.13+ to override options when using FetchContent / add_subdirectory.
+    cmake_policy(SET CMP0077 NEW)
+endif ()
+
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -24,6 +24,11 @@ include( GNUInstallDirs )
 
 get_directory_property( has_parent PARENT_DIRECTORY )
 
+if (POLICY CMP0077)
```
TheStormN, line 27: This makes it compatible with older CMake versions which do not support this policy.

TheStormN:
@HowardHinnant I know you are not promoting CMake support, but in general it does work and tbh I don't think that just copying over some files into a project is a nice way for 3rd party dependency management.

My ask is to consider merging this PR, so my(and other people) life can become a bit easier. :)

PR #814, opened 2024-02-17
Fixed curl linkage when building date-tz

BestITUserEUW:


BestITUserEUW:
Fails due to unresolved external symbols of curl. Happens with GCC-11.4.0 and MSVC 19.34.31937.0

basilgello:
@BestITUserEUW Is this PR still relevant for your workflow? I can not reproduce the behavior on Debian containers.

BestITUserEUW:
@basilgello  I can try and give you a reproducable example in the next few days. Right it's not important as i am using a patch.

BestITUserEUW:
Reopened this issue as it seems that other people have this problem too: #856

PR #831, opened 2024-06-24
Fix UWP build

DDoS:
These changes come from the [Vcpkg package update](https://github.com/microsoft/vcpkg/pull/39465). I believe these would be better to have in the library itself.

### DIFF
--- a/include/date/date.h
+++ b/include/date/date.h
@@ -81,7 +81,7 @@
 #ifdef _MSC_VER
 #   pragma warning(push)
 // warning C4127: conditional expression is constant
-#   pragma warning(disable : 4127)
+#   pragma warning(disable : 4127 4996)
 #endif
 
 namespace date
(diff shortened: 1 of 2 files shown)


include/date/date.h
```diff
@@ -81,7 +81,7 @@
 #ifdef _MSC_VER
 #   pragma warning(push)
 // warning C4127: conditional expression is constant
-#   pragma warning(disable : 4127)
+#   pragma warning(disable : 4127 4996)
```
TheStormN, line 84: What is the code which causes 4996?

DDoS: I just copied this from the original UWP support vcpkg patch. Looks like it's originally from [here](https://github.com/microsoft/vcpkg/pull/8151#issuecomment-531175393). I don't have a Windows dev environment, so I can't test these changes directly, but I know it fixed the UWP CI tests in vcpkg.

PR #865, opened 2025-02-13
fix(date.h): uninitialized variables error

utkarshmsr3:
```
error: ‘tH’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
```
This could have been handled with https://github.com/HowardHinnant/date/pull/698


HowardHinnant:
Looks like a gcc bug to me.  I can't find a path of use of the variables if they are uninitialized.  The initialized state is protected by the stream fail checks.

Issue #872, opened 2025-03-11
Usage of `rs{}` vs `ru{}`


DavisVaughan:
When looking at #871 I was also closely looking at current usage of `rs{}`. It seems like many of these places should actually be `ru{}`? i.e. I _think_ the places highlighted below never expect `+` or `-` signs, and the only places that should actually use `rs{}` are when parsing years or iso based years - which I think would also let you make changes to `read_signed()` with even more confidence, knowing that it only affects year parsing.

- `%c` reading day of the month https://github.com/HowardHinnant/date/blob/d18e8b1653e0ab1f583dcea22f007e12f69e497b/include/date/date.h#L6787
- `%x` reading the two digit year https://github.com/HowardHinnant/date/blob/d18e8b1653e0ab1f583dcea22f007e12f69e497b/include/date/date.h#L6837
- `%C` reading the century https://github.com/HowardHinnant/date/blob/d18e8b1653e0ab1f583dcea22f007e12f69e497b/include/date/date.h#L6900
- `%D` reading the two digit year https://github.com/HowardHinnant/date/blob/d18e8b1653e0ab1f583dcea22f007e12f69e497b/include/date/date.h#L6933
- `%d` and `%e` reading the day of the month https://github.com/HowardHinnant/date/blob/d18e8b1653e0ab1f583dcea22f007e12f69e497b/include/date/date.h#L6981
- `%I` reading the 12 based hour (notably `%H` uses `ru{}` already!) https://github.com/HowardHinnant/date/blob/d18e8b1653e0ab1f583dcea22f007e12f69e497b/include/date/date.h#L7045-L7047
- `%m` reading the month of the year https://github.com/HowardHinnant/date/blob/d18e8b1653e0ab1f583dcea22f007e12f69e497b/include/date/date.h#L7119
- `%z` reading the hours part of the offset (b/c you've already dealt with a leading `+` or `-` up front here https://github.com/HowardHinnant/date/blob/d18e8b1653e0ab1f583dcea22f007e12f69e497b/include/date/date.h#L7529-L7559

Issue #871, opened 2025-03-11
Clarification on `read_signed()` and usage on negative years with `%Y`


DavisVaughan:
I'm working on resyncing my library with the latest version of date and I noticed these two commits:

https://github.com/HowardHinnant/date/commit/ab37c362e35267d6dee02cb47760f9e9c669d3be

https://github.com/HowardHinnant/date/commit/5d225951ff24e6802d5026cb85d5b5a95b1a2476

The change to `read_signed()` seems pretty meaningful to me, as this is used in `%Y` parsing. IIUC it means that `-1000-01-01` no longer parses out of the box with `%Y-%m-%d`. You now need `%5Y-%m-%d` to parse this (same with `⁠%F` needing `⁠%5F`).

Also, it seems like `%c` has no way to parse negative years now.

In concrete terms, this test I just wrote used to pass but now does not:

```cpp
void
test_Y()
{
    using namespace date;
    using namespace std::chrono;
    {
        std::istringstream in{"-1000-02-03"};
        sys_days tp;
        in >> parse("%Y-%m-%d", tp);
        assert(!in.fail());
        assert(!in.bad());
        assert(tp == -1000_y/02/03);
    }
}
```

Was that meaningful? If so, here's a test that may be worth adding, there don't seem to be any for parsing negative years right now. It also may be worth calling this case out explicitly in the docs (I realize it says that it uses a width of 4 by default, but the negative sign adds some ambiguity around that IMO)

```cpp
void
test_Y()
{
    using namespace date;
    using namespace std::chrono;
    {
        std::istringstream in{"-1000-02-03"};
        sys_days tp;
        in >> parse("%Y-%m-%d", tp);
        assert(in.fail());
        assert(in.bad());
    }
    {
        std::istringstream in{"-1000-02-03"};
        sys_days tp;
        in >> parse("%5Y-%m-%d", tp);
        assert(!in.fail());
        assert(!in.bad());
        assert(tp == -1000_y/02/03);
    }
}
```

DavisVaughan:
@HowardHinnant is it possible you missed this one when looking at the other pr yesterday? (no rush if you did see this, you're just always so fast to respond!)

DavisVaughan:
[...]

PR #873, opened 2025-03-11
Use `ru{}` in favor of `rs{}` where it makes sense

DavisVaughan:
Closes https://github.com/HowardHinnant/date/issues/872

I've swapped out `rs` for `ru` everywhere I mentioned in #872 except for in `%C` where apparently century can be meaningfully negative. (Normally I'd wait for you to comment before doing a PR but it seemed like a POC was an easier way to see if I was missing anything with my idea)

I also went ahead and added some extra tests related to this. The majority of the new tests would not have passed before this change (meaning they would have previously performed a successful parse in some weird cases).

The test for a negative year in `%c` failing is dependent on the decision of #871, but reflects the current state of things.


HowardHinnant:
Thanks, nice job!

PR #877, opened 2025-03-18
CMake: Fix Apple Mobile platforms detection

### LINKED ISSUE #876: Apple mobile platforms other than iOS fails to compile with linker error
Apple mobile platforms like visionOS, watchOS and tvOS fails to compile, because the currect CMake file check is adding the apple mobile platform file only when iOS is detected. The if() checks should be updated to properly compile for the other apple platforms too.

P.S. I will submit a PR soon.

TheStormN:
Fixes #876


TheStormN:
@HowardHinnant could you please take a look at this simple fix

PR #884, opened 2025-06-27
compatible for c++11 but GCC <

aengusjiang:
a little change to support c++11 but GCC < 5, such as centos7.9 gcc4.8.5

### DIFF
--- a/include/date/date.h
+++ b/include/date/date.h
@@ -84,6 +84,12 @@
 #   pragma warning(disable : 4127)
 #endif
 
+#if (defined(__GNUC__) && __GNUC__ < 5) || defined(_MSC_VER)
+#  define OPERATOR_LITERAL(suffix) operator"" suffix
+#else
+#  define OPERATOR_LITERAL(suffix) operator""suffix
+#endif
+
 namespace date
 {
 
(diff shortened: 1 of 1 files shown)


include/date/date.h
```diff
@@ -84,6 +84,12 @@
 #   pragma warning(disable : 4127)
 #endif
 
+#if (defined(__GNUC__) && __GNUC__ < 5) || defined(_MSC_VER)
```
HowardHinnant, line 87: Why " || defined(_MSC_VER)"?

aengusjiang:
#if (defined(__GNUC__) && !defined(__clang__) && __GNUC__ < 5)
#  define OPERATOR_LITERAL(suffix) operator"" suffix
#else
#  define OPERATOR_LITERAL(suffix) operator""suffix
#endif

aengusjiang:
may  this can help you 
#if (defined(__GNUC__) && !defined(__clang__) && __GNUC__ < 5)
#  define OPERATOR_LITERAL(suffix) operator"" suffix
#else
#  define OPERATOR_LITERAL(suffix) operator""suffix
#endif

Issue #883, opened 2025-06-18
The Windows version just assumes 7z is installed


bopt-woba:
The Windows version has a silent dependency on a 7z installation. There is no way around the default discovery mechanism, which only looks in the registry and the program files directory.

- I cannot use a custom path
- 7z will not be looked up in %PATH%
- 7z is not necessary at all

The funny thing is, that tar is available on any Windows starting with Windows 10 build 17063 (December 2017), so just using that would be easier. Windows systems before that should not be used anymore anyway.

So please use tar (includes automatic .gz decompression), which should reduce the code complexity while removing gratuitous dependencies. There is no need for custom software installations on Windows. The following will suffice:

```bash
mkdir tzdata
tar xf tzdata2025b.tar.gz -C tzdata
```

Issue #25, opened 2015-11-04
constexpr support in VS 2015


HowardHinnant:
Currently `constexpr` support is turned off in VS-2015 due to the lack of ability to forward declare `constexpr` functions.

This issue is to serve as a reminder to re-check `constexpr` settings with VS-2015 Update 2 (unsure of the timeline for that release).  I have reason to believe we can start turning `constexpr` on for VS at that time (hopefully at least the C++11 parts).

davidhunter22:
I did try 2015 Update 1 to see if there was any progress at all. That actually led to a compiler crash, see https://connect.microsoft.com/VisualStudio/feedback/details/1980909

philsquared:
Update 2 has been out for a while now (and Update 3 is available in RC!). Has anyone tried this yet?

gmcode:
hi phil*phil, I've tried update 2, this library works with it. I haven't tried update 3 yet as I was waiting for final release. I will be sure to install update 3 as soon as it is released in final form.

philsquared:
Thanks @gmcode  (btw, that's +ᵖʰⁱˡ(phil) to you ;-) ).
When you say, "this library works with it" do you specifically mean the `constexpr` support?

gmcode:
Hi Phil, I meant this library works with cl.exe (at least VS2014 / MSC_VER 19.00.23918) but not with full constexpr support because that version of VS doesn't have full support as you know. VS Update 3 has better constexpr support I believe but to what level I'm not sure. I haven't tried Update 3 with this library. I suspect a patch will be needed at that time, but it should be a small one (if you're using Update 3 RC try it, patches welcome!) as full constexpr does work in the clang variants and g++ 6 including when they run on Windows.

philsquared:
AFAIK update 2 had more `constexpr` support the update 1 (which had more than GM). For some values of "more", of course :-) Since I don't know what those values are (they never seem to be spelled out!) I was interested to see whether update 2 had enough. It sounds like not.
I'm still trying to get our code base working with update 2 (well 2015 in general). Sadly I don't have the bandwidth to try update 3 yet while it's still in beta. I'd certainly be keen to hear from anyone else if they've tried it, though.
We have an aging Date/DateTime abstraction (wrapping Boost.DateTime) that I'm looking to replace with `std::chrono` and this library in the near future. constexpr would be a nice-to-have but is not a pre-req for our adoption so it's just general interest at this stage.

gmcode:
[...]

PR #895, opened 2025-09-17
Use built-in tar instead of 7zip on windows.

deivse:
This removes the implied 7zip dependency in favour of using `tar` which has been available by default on Windows for a while now, as noted in #883 (See https://techcommunity.microsoft.com/blog/containers/tar-and-curl-come-to-windows/382409).

I needed to change this for my own use, so decided to contribute, hopefully this is acceptable. Imho, requiring windows versions released after 2017/18 is better than requiring 7zip. Alternatively, this can be reworked to try to find both `7zip` and `tar` and decide which one to use, but imo that's overkill considering this is supposed to be replaced by the stdlib implementation (any time now, Apple...)

### DIFF
--- a/src/tz.cpp
+++ b/src/tz.cpp
@@ -3421,50 +3421,13 @@ move_file(const std::string& from, const std::string& to)
 #    endif // !USE_SHELL_API
 }
 
-// Usually something like "c:\Program Files".
-static
-std::string
-get_program_folder()
-{
-    return get_known_folder(FOLDERID_ProgramFiles);
-}
-
-// Note folder can and usually does contain spaces.
+// Note folder can contain spaces.
 static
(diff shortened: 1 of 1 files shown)


pr0m1th3as:
@HowardHinnant is there an incentive to merge this PR? I am using `date` in the [datatypes](https://github.com/pr0m1th3as/datatypes) package for GNU Octave in order to support the `datetime` classdef. I have received a few complaints from Windows users getting errors [issue #25](https://github.com/pr0m1th3as/datatypes/issues/25), which is related to the missing 7-zip dependency on their systems, and I am wondering whether I should apply this patch to my local sources or wait until is merged and pull the latest `date` into my project. Thanks.

PR #919, opened 2026-08-06
fix: signed-int-overflow in `date::from_stream` width parsing

### LINKED ISSUE #918: Signed integer overflow in `date::from_stream` when parsing `%` width specifier
## Summary

When the `date` library parses a format string, `from_stream` accumulates a width from consecutive digits following `%` using `width = 10*width + digit` (`include/date/date.h:7640`) with no overflow check. A format string containing enough consecutive digits (e.g. `%5555555555555555U`) drives the `int` accumulator past `INT_MAX`, which is signed integer overflow — undefined behavior that a UBSan-instrumented build aborts on.

## Version

```bash
$ git describe --tags
v3.0.4-10-g7875f43
```

## Description

`from_stream` (`include/date/date.h`) parses `strftime`/`strptime`-style format strings. When it encounters `%` followed by digits, it interprets those digits as a width specifier, accumulated into a plain `int`:

```cpp
// include/date/date.h:6605
auto width = -1;
...
// include/date/date.h:7636-7641
if (width == -1 && modified == CharT{} && '0' <= *fmt && *fmt <= '9')
{
    width = static_cast<char>(*fmt) - '0';
    while ('0' <= fmt[1] && fmt[1] <= '9')
        width = 10*width + static_cast<char>(*++fmt) - '0';  // UB on overflow
}
```

The loop has no overflow guard. With sixteen consecutive `'5'` digits the accumulator reaches `555555555` and the next `10 * 555555555 = 5555555550` exceeds `INT_MAX` (2147483647). Signed integer overflow is undefined behavior per C++ [expr.pre]/4; UndefinedBehaviorSanitizer flags and aborts on it:

```
date.h:7640:39: runtime error: signed integer overflow: 10 * 555555555 cannot be represented in type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior date.h:7640:39
```

`width` is `int` (deduced from `auto width = -1;`, line 6605), and the accumulated value is later consumed by the various `read(is, …, width == -1 ? N : static_cast<unsigned>(width))` call sites scattered through the same function (e.g. lines 6908, 6963, 6989). So an overflowed `width` also propagates into the read-width argument of every field parser that consults it.

This is a library bug, not API misuse:

- The PoC uses the documented public API directly: construct an `std::istringstream`, default-construct a `date::year_month_day`, call `date::from_stream(is, fmt, ymd)`. No undocumented entry point, no manual field/state manipulation.
- There is no API contract restricting the length or content of the format string. `from_stream` accepts an arbitrary `const CharT* fmt` and is documented to parse `%`-prefixed specifiers; a caller passing an attacker- or user-supplied format string is within the documented usage. The library is responsible for handling arbitrary format strings without invoking UB.
- The crash occurs entirely inside the library's format-string parser; no pointer the caller supplies is null or otherwise invalid. The overflow is generated internally by the unchecked `10*width + digit` accumulation.

### Trigger

The minimized trigger is the format literal `"%5555555555555555U"`: `%`, sixteen `'5'` digits to overflow the accumulator, and a trailing `U` (a recognized two-digit-year specifier) so the format scan reaches the width-parsing branch rather than treating the digits as part of a different directive. Any input stream content suffices (the stream body is never reached); a stream holding `"2024-01-01"` is used purely so the `istream` is in a good state.

## PoC Code

```cpp
#include <date/date.h>
#include <sstream>

// Reproduces signed-integer overflow in date::from_stream (date.h:7640):
// the width accumulator `width = 10*width + digit` overflows `int` when the
// format string holds enough consecutive digits after '%'.
int main() {
    std::istringstream iss("2024-01-01");
    date::year_month_day ymd{};
    date::from_stream(iss, "%5555555555555555U", ymd);
    return 0;
}
```

## Stack Trace

```
runtime error: signed integer overflow: 10 * 555555555 cannot be represented in type 'int'
    #0 0x629a8ec71968 in std::basic_istream<char, std::char_traits<char>>& date::from_stream<char, std::char_traits<char>, std::chrono::duration<long, std::ratio<1l, 1l>>, std::allocator<char>>(std::basic_istream<char, std::char_traits<char>>&, char const*, date::fields<std::chrono::duration<long, std::ratio<1l, 1l>>>&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>>*, std::chrono::duration<long, std::ratio<60l, 1l>>*) /date.h:7640:39
    #1 0x629a8ec5fb80 in std::basic_istream<char, std::char_traits<char>>& date::from_stream<char, std::char_traits<char>, std::allocator<char>>(std::basic_istream<char, std::char_traits<char>>&, char const*, date::year_month_day&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>>*, std::chrono::duration<long, std::ratio<60l, 1l>>*) /date.h:8024:5
    #2 0x629a8ec5f66a in main /poc.cpp:10:5
    #3 0x7f1f534e51c9 in __libc_start_call_main csu/../sysdeps/nptl/libc_start_call_main.h:58:16
    #4 0x7f1f534e528a in __libc_start_main csu/../csu/libc-start.c:360:3
    #5 0x629a8eb74674 in _start (/poc+0x3e674) (BuildId: 708ba1ac4cf9f1cc216db0727a231565d73972df)

SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /date.h:7640:39 
```

## Reproduction Step

```bash
# 1. Checkout the date library at the pinned commit
git clone https://github.com/HowardHinnant/date.git
cd date
git checkout 7875f43aa9288d176191c83e8284d61acdd5ab7c

# 2. Build the library with sanitizer flags (static, ASan+UBSan)
export CC=clang CXX=clang++
export CFLAGS="-fsanitize=address,undefined -fno-sanitize=function -g -O0 -fno-omit-frame-pointer"
export CXXFLAGS="$CFLAGS"
cmake -B build/sanitizer \
      -DCMAKE_INSTALL_PREFIX="$PWD/build/sanitizer" \
      -DCMAKE_C_COMPILER="$CC" -DCMAKE_CXX_COMPILER="$CXX" \
      -DCMAKE_C_FLAGS="$CFLAGS" -DCMAKE_CXX_FLAGS="$CXXFLAGS" \
      -DBUILD_SHARED_LIBS=OFF \
      -DBUILD_TZ_LIB=ON \
      -DUSE_SYSTEM_TZ_DB=ON \
      -DMANUAL_TZ_DB=OFF \
      -DENABLE_DATE_TESTING=OFF
cmake --build build/sanitizer -j$(nproc)
cmake --install build/sanitizer
# Ensure headers are present where the PoC include path expects them
mkdir -p build/sanitizer/include/date
cp include/date/*.h build/sanitizer/include/date/

# 3. Build the standalone PoC. It is a plain main() that calls the public
#    date::from_stream API with the trigger format literal inline — no file
#    input, no argv, no libFuzzer, no fuzzer RNG.
clang++ -g -O0 -fsanitize=address,undefined \
    -I"$PWD/build/sanitizer/include" \
    poc.cpp -o poc \
    -Wl,--start-group $PWD/build/sanitizer/lib/lib*.a -Wl,--end-group

# 4. Run
./poc
```

## Impact

Triggering requires only that a caller pass an attacker-controllable format string to `date::from_stream` — the normal use of the parsing API when processing untrusted input. The observable effect is a process abort under a sanitizer build; on a non-sanitized build the overflow is silent (the overflowed `int` wraps per implementation-defined behavior and execution continues with a corrupted width), so the practical severity is low (UB hygiene rather than memory corruption). Exploitation beyond a sanitizer-only abort was not evaluated. The format string is an ordinary value, so input validation does not inherently catch it unless the caller separately caps format-string digit runs.

## Submission Statement

This report was produced by FuzzAnything's AI-assisted library fuzzer and manually verified by a team member. We reviewed the harness against the upstream API documentation — call order, parameters, memory ownership — and found no API misuse.
> Signed-off-by: FuzzAnything <fuzzanything@gmail.com>

yvonnelxxxx:
This PR attempts to fix [#918](https://github.com/HowardHinnant/date/issues/918#issue-5077552571)

## Problem

`date::from_stream` parses `strftime`/`strptime`-style format strings. When it encounters `%` followed by digits, it interprets those digits as a width specifier, accumulated into a plain `int` (`auto width = -1;`, line 6605):

```cpp
// include/date/date.h:7636-7641
if (width == -1 && modified == CharT{} && '0' <= *fmt && *fmt <= '9')
{
    width = static_cast<char>(*fmt) - '0';
    while ('0' <= fmt[1] && fmt[1] <= '9')
        width = 10*width + static_cast<char>(*++fmt) - '0';  // UB on overflow
}
```

The loop has no overflow guard. With sixteen consecutive `'5'` digits the accumulator reaches `555555555` and the next `10 * 555555555 = 5555555550` exceeds `INT_MAX` (2147483647). Signed integer overflow is undefined behavior per C++ [expr.pre]/4; UndefinedBehaviorSanitizer flags and aborts on it:

```
date.h:7640:39: runtime error: signed integer overflow: 10 * 555555555 cannot be represented in type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior date.h:7640:39
```

The overflowed `width` also propagates into the read-width argument of every field parser that consults it, via `width == -1 ? N : static_cast<unsigned>(width)` (e.g. lines 6908, 6963, 6989).

This is a library bug, not API misuse: the PoC uses the documented public API directly (`std::istringstream` + default-constructed `date::year_month_day` + `date::from_stream(is, fmt, ymd)`), and there is no API contract restricting the length or content of the format string. The library is responsible for handling arbitrary format strings without invoking UB.

## Fix

Guard the multiply before it can overflow, breaking out of the digit loop so `width` stays at the largest in-range value accumulated so far:

```cpp
// include/date/date.h — width loop
if (width == -1 && modified == CharT{} && '0' <= *fmt && *fmt <= '9')
{
    width = static_cast<char>(*fmt) - '0';
    while ('0' <= fmt[1] && fmt[1] <= '9')
    {
        const int digit = static_cast<char>(*++fmt) - '0';
        if (width > (numeric_limits<int>::max() - digit) / 10)
            break;  // prevent signed overflow
        width = 10*width + digit;
    }
}
```

`numeric_limits` is already in use elsewhere in the file, so no new include is required. On overflow the directive proceeds with a (large) in-range width — equivalent to a caller legitimately writing e.g. `%999999999Y` — which reads as many digits as the stream offers: bounded, no UB, no infinite loop.

### Sanitizer rebuild

Rebuilt the library with AddressSanitizer and UndefinedBehaviorSanitizer to
confirm the overflow is gone and no UB remains:

```bash
export CC=clang CXX=clang++
export CFLAGS="-fsanitize=address,undefined -fno-sanitize=function -fsanitize-address-use-after-scope -g -O0 -fstandalone-debug -fPIC -fno-inline -fno-omit-frame-pointer -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION"
export CXXFLAGS="$CFLAGS"
cmake -S "$SRC" -B /tmp/build_tmp \
      -DCMAKE_INSTALL_PREFIX="$WORK" \
      -DCMAKE_C_COMPILER="$CC" -DCMAKE_CXX_COMPILER="$CXX" \
      -DCMAKE_C_FLAGS="$CFLAGS -g" -DCMAKE_CXX_FLAGS="$CXXFLAGS -g" \
      -DBUILD_SHARED_LIBS=OFF -DBUILD_TZ_LIB=ON \
      -DUSE_SYSTEM_TZ_DB=ON -DMANUAL_TZ_DB=OFF \
      -DENABLE_DATE_TESTING=OFF -DENABLE_DATE_INSTALL=ON
cmake --build /tmp/build_tmp -j$(nproc)
cmake --install /tmp/build_tmp
```

### Before fix

```
date.h:7640:39: runtime error: signed integer overflow:
10 * 555555555 cannot be represented in type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior date.h:7640:39   (exit 1)
```

### After fix

(clean exit, code 0 — no sanitizer errors)

### PoC test

```bash
clang++ -g -O0 -fsanitize=address,undefined \
    -I"$WORK/include" poc.cpp -o poc \
    -Wl,--start-group "$WORK"/lib/lib*.a -Wl,--end-group
./poc   # "%5555555555555555U" → previously tripped UBSan; now exits 0 clean
```

### Regression check

A/B compared the original (unpatched) and patched headers across normal parsing and bounded-width specifiers:

| format              | original header | patched header |
|---------------------|-----------------|----------------|
| `%F` → `2024-03-15` | `2024-3-15`, fail=0 | `2024-3-15`, fail=0 |
| `%Y` / `%4Y`        | year=1999, fail=1 | year=1999, fail=1 (identical) |
| `%5555555555555555U`| UBSan abort (exit 1) | clean exit 0 |

The `%Y`/`%4Y` `fail=1` is pre-existing behavior (a single `%Y` specifier against a stream containing only `"2024"` does not satisfy `year_month_day` validation) and is unchanged by the fix — confirming normal parsing is byte-for-byte identical; the fix is strictly additive, only preventing the overflow.

### DIFF
--- a/include/date/date.h
+++ b/include/date/date.h
@@ -7637,7 +7637,12 @@ from_stream(std::basic_istream<CharT, Traits>& is, const CharT* fmt,
                     {
                         width = static_cast<char>(*fmt) - '0';
                         while ('0' <= fmt[1] && fmt[1] <= '9')
-                            width = 10*width + static_cast<char>(*++fmt) - '0';
+                        {
+                            const int digit = static_cast<char>(*++fmt) - '0';
+                            if (width > (numeric_limits<int>::max() - digit) / 10)
+                                break;
+                            width = 10*width + digit;
+                        }
                     }
                     else
                     {


HowardHinnant:
Thanks!  Fixed by limiting width to two digits.

Issue #921, opened 2026-08-24
Incorrect time zone name displayed by `date::format("%c")`


cschreib-ibex:
Hello, and thanks for putting together the date/tz library! 

I'm encountered an issue with `date::format()` when using the `"%c"` specifier: 
```c++
auto time = make_zoned(locate_zone("Asia/Bangkok"), local_days{Monday[1] / May / 2016} + 9h);
std::cout << format(std::locale("en_GB.UTF-8"), "%c", time) << std::endl;
```

On my machine (Linux, Ubuntu 22.04) this prints:
```
Mon 02 May 2016 09:00:00 GMT
```

Notice how the string includes an incorrect "GMT" time zone.

Looking at the source code of `date`, I traced it back to [the use of the `std::tm` struct](https://github.com/HowardHinnant/date/blob/4bb501c01b0fad72a94adce9936fbfe155f155d7/include/date/date.h#L5146-L5167), and not setting the `tm_zone` member. In this instance, the locale falls back to using the `TZ` environment variable, which (in this case) is set to a different time zone (GMT).

Here's another code snippet that allows reproducing the problem with just the STL:
```c++
#include <iostream>
#include <locale>
#include <sstream>

int main() {
  std::tm tm;
  tm.tm_year = 109;
  tm.tm_mday = 5;
  tm.tm_mon = 8;
  tm.tm_wday = 1;
  tm.tm_yday = 205;
  tm.tm_sec = 55;
  tm.tm_min = 12;
  tm.tm_hour = 3;

  // tm.tm_zone = "CET";

  std::ostringstream os;
  os.imbue(std::locale("en_GB.UTF-8"));

  auto &facet = std::use_facet<std::time_put<char>>(os.getloc());

  char f[3] = {'%', 'c'};
  auto fe = std::begin(f) + 2;

  facet.put(os, os, os.fill(), &tm, f, fe);

  std::cout << os.str() << std::endl;

  return 0;
}
```

On my machine, this prints:
```
Mon 05 Sep 2009 03:12:55 GMT
```

Setting `TZ="Asia/Bankok"` before running the code, this prints:
```
Mon 05 Sep 2009 03:12:55 Asia
```

Un-commenting the line that sets `tm_zone`, this prints:
```
Mon 05 Sep 2009 03:12:55 CET
```

[...]

PR #922, opened 2026-08-26
Fix wrong time zone displayed with %c on some locales.

cschreib-ibex:
See motivation and discussion in https://github.com/HowardHinnant/date/issues/921.

On Ubuntu linux, some locales include the time zone in the `%c` format. Previously this would always display the system time zone, even if the time point being formatted included another time zone. This fix uses the non-standard `tm_zone` field of the `std::tm` struct to pass through the input time zone, which solves the problem.

Not all platforms offer this field, and some platforms offer it but ignore it. There is not much we can do to help in these cases.

A heuristic is included to guess whether the platform supports this field, which aims to catch glibc, android, and muslc, but not BSD or MacOS (where the field is non-const and unclear whether it is allowed to set it manually). In case the heuristic is wrong, a preprocessor and CMake variable can be used to disable it.

In #921 I had suggested that we should error if `%c` is used for a time point that does not include time zone information (e.g., `local_time`), since we can't prevent the locale from appending the time zone in these cases. But that is a bit mean: most locales and most platforms will not do that. Furthermore, it is allowed in `std::chrono`.


include/date/date.h
```diff
@@ -157,6 +157,14 @@ namespace date
 #  endif
 #endif  // HAS_VOID_T
 
+#ifndef HAS_TM_ZONE
+#  if defined(__GLIBC__) || defined(__ANDROID__) || defined(__GNUC__)
```
HowardHinnant: macOS defaults `HAS_TM_ZONE` to 1 causing this error:

```
error: assigning to 'char *' from 'const value_type *' (aka 'const char *') discards qualifiers
 5177 |                     tm.tm_zone = abbrev != nullptr ? abbrev->c_str() : nullptr;
      |                                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
Changing this line to:

```
#  if defined(__GLIBC__) || defined(__ANDROID__) || (defined(__GNUC__) && __GNUC__ > 4)
```

will fix it.

Explanation:  When Apple replaced GCC with clang they were so paranoid about backwards compatibility that clang defines `__GNUC__` to 4, which is what it was in the version of gcc they replaced.  And it has stayed 4 ever since.

HowardHinnant (changes requested):

include/date/date.h
cschreib-ibex: Ah I didn't realise this was why the `> 4` was needed, thanks for the background. I have included this now.

